PR #289, opened 2015-01-31
cdrtools devel 3.01a27 for Tigerbrew

fbrosson:
I'm not sure if this is the right way to update cdrtools to 3.01a27 for Tigerbrew. Sorry if it is not.
BTW, I've added a "test" target similar to the one added to cdrtools 3.01a26 for Homebrew.

### DIFF
--- a/Library/Formula/cdrtools.rb
+++ b/Library/Formula/cdrtools.rb
@@ -14,9 +14,9 @@ class Cdrtools < Formula
   end
 
   devel do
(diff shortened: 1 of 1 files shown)


mistydemeo:
Thank you for the submission! This ended up getting updated when I merged in the most recent set of Homebrew changes.

fbrosson:
Hello Misty De Meo! I wish I had acknowledged your kind reply earlier.
I noticed you have recently changed the picture of your GitHub profile. I don't remeber what the previous one looked like, but I do like the new one ;-) BTW, I also like your website and the good old PowerPC-based Macs!

PR #297, opened 2015-02-03
Adds Tigerbrew flex as dependency for gstreamer

dcwar:
This adds flex as a dependency for gstreamer in Tigerbrew.


mistydemeo:
Thank you very much! Merged.

PR #311, opened 2015-03-30
cdrtools devel 3.01a28

fbrosson:
This edit should make Tigerbrew's formula for cdrtools in sync with that in Homebrew.
The main (tiny) changes are:
1) Upgrade from 3..01a27 to 3.01a28 (I also dropped the useless tilde I had added in the devel version a few months ago when I added the devel block).
2) Moved the bottle block outside of the stable block.
3) Slightly improved the test block (for mkisofs).
Thanks again (and sorry if I'm not contributing the right way).


mistydemeo:
Thank you! I've merged in the latest Homebrew changes, so I think this should be addressed by that.

fbrosson:
:wink:

PR #317, opened 2015-04-18
git: Fix stray whitespace

chdiza:


chdiza:
These two spaces snuck in recently.

mistydemeo:
Thank you!

PR #319, opened 2015-04-19
cmake: Use --no-system-curl

chdiza:


chdiza:
Not sure if this is the right thing to do, but I found that I could not otherwise get cmake 3.2.1 to build on TigerPPC.  I tried with `curl` force-linked; I tried with stdenv; I tried with gcc42 and gcc49; It would always flake out with some error mentioning `curl` in it somewhere.

Also, I'm not sure if this is necessary on Leo.

mistydemeo:
Does it pick up on a Tigerbrew-installed curl without this if you add it as a conditional dependency?

chdiza:
I dunno, haven't tried.  It takes ages to build on my 12-inch G4.  I'll try when I get a chance.

Incidentally: why on earth does a build system need curl?

chdiza:
> Does it pick up on a Tigerbrew-installed curl without this if you add it as a conditional dependency?

Experiment reveals that the flag `--system-curl` absolutely MUST NOT be passed, even if `depends_on "curl"` is added.

I found that with a dep on curl explicitly declared, one could either pass `--no-system-curl` or just outright delete the `--system-curl` and not replace it, and then cmake would build.

I favor passing `--no-system-curl` just to be on the safe side, but then again, I really have no idea what I'm talking about :)

I think this problem has been around for a long time, but it only just now got worse.  Until recently, one had to make sure that a brewed curl was NOT picked up, lest brew barf.  But they must have tweaked something that rendered system curl on Tiger unworthy.

chdiza:
I don't know whether this is needed on Leopard.

mistydemeo:
I tried to convince it to use the Homebrew curl but, uh, it didn't work in a weird way:

```
Linking CXX executable ../bin/ccmake
/usr/bin/ld: warning can't open dynamic library: /Developer/SDKs/MacOSX10.4u.sdk/usr/local/opt/openssl/lib/libssl.1.0.0.dylib referenced from: /usr/local/opt/curl/lib/libcurl.dylib (checking for undefined symbols may be affected) (No such file or directory, errno = 2)
/usr/bin/ld: warning can't open dynamic library: /Developer/SDKs/MacOSX10.4u.sdk/usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib referenced from: /usr/local/opt/curl/lib/libcurl.dylib (checking for undefined symbols may be affected) (No such file or directory, errno = 2)
```

chdiza:
Hmm.  I probably built it in `stdenv`, but now I can't remember.

mistydemeo:
I completely missed your comment about changing the `curl` flags; retesting with those adjusted.

mistydemeo:
Pushed a tweaked version of this, thank you very much!

PR #321, opened 2015-04-25
ruby 2.2.2

### LINKED ISSUE #318: ruby is unbrewable on PPC
Ruby 2.2.1 and 2.2.2 refuse to brew on TigerPPC.  It doens't make a difference whether I use stdenv or superenv, or whether I use the official formula or a custom one.

No matter what, I get this:
```
compiling newline.c
compiling ./missing/setproctitle.c
compiling ./missing/alloca.c
compiling dmyext.c
compiling dmyenc.c
linking miniruby
/usr/bin/ld: Undefined symbols:
_fgetattrlist
collect2: ld returned 1 exit status
make: *** [miniruby] Error 1
^C/usr/local/Library/Homebrew/utils/popen.rb:16:in `read'
````
Occasionally the error message is ever so slightly different, e.g., if I try adding a `depends_on "ld64"`, but it's the same thing: `ld` can't find some symbols.[1]

Ruby 2.2.0 has no such problem.

[1] If I add an ld64 dep, I get this:
```
compiling newline.c
compiling ./missing/setproctitle.c
compiling ./missing/alloca.c
compiling dmyext.c
compiling dmyenc.c
linking miniruby
Undefined symbols:
  "_fgetattrlist", referenced from:
      _dir_each in dir.o
      _glob_helper in dir.o
ld: symbol(s) not found
collect2: ld returned 1 exit status
make: *** [miniruby] Error 1
```

chdiza:
Also, patch to fix #318.  Remove it at next point release.

### DIFF
--- a/Library/Formula/ruby.rb
+++ b/Library/Formula/ruby.rb
@@ -1,7 +1,7 @@
 class Ruby < Formula
   homepage "https://www.ruby-lang.org/"
-  url "http://cache.ruby-lang.org/pub/ruby/2.2/ruby-2.2.1.tar.bz2"
-  sha256 "4e5676073246b7ade207be3e80a930567a88100513591a0f19fc38e247370065"
+  url "http://cache.ruby-lang.org/pub/ruby/2.2/ruby-2.2.2.tar.bz2"
+  sha256 "f3b8ffa6089820ee5bdc289567d365e5748d4170e8aa246d2ea6576f24796535"
 
   bottle do
   end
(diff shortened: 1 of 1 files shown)


chdiza:
Note that the patch applies with offsets.  I'm embarrassed to admit that I don't know exactly how to modify the line numbers in the diff.

mistydemeo:
Hm, the patch changes configure.in, so it triggers an `autoreconf` (which fails to to the lack of a suitable version).

mistydemeo:
I was able to remove the `configure.ac` patch entirely; looks like the prebuilt configure script in 2.2.2 already includes that change. Builds great with the rest of the patch. Thanks!

PR #327, opened 2015-05-15
Move some monkeypatches into extend/tiger.rb

jacknagel:
These are only required on ruby < 1.8.7, so if we move them here I can remove them upstream.

### DIFF
--- a/Library/Homebrew/extend/leopard.rb
+++ b/Library/Homebrew/extend/leopard.rb
@@ -1,0 +1,35 @@
+module Enumerable
+  def group_by
+    inject({}) do |h, e|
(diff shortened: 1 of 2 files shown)


mistydemeo:
Hm, I think I need to reorganize these or otherwise find a better place for them. `extend/tiger` is only meant to be included in 1.8.2. If you want to remove the monkeypatches though I can just revert that commit the next time I merge in changes.

jacknagel:
I can stick them in a file for <= 1.8.6 if that would make it easier.

jacknagel:
Pushed an alternate approach.

mistydemeo:
Merged, thanks!

PR #342, opened 2015-06-25
Add dependency on `make` package.

qu1j0t3:
This is needed in Tigerbrew because Chicken no longer builds with the system make 3.81.

What I would *really* like is to be able to dep on `make`, then use the Cellar'd make binaries without requiring `brew link`. This would allow the brew to work handsfree and not require an otherwise unnecessary change to environment. But I don't know if that's possible.

### DIFF
--- a/Library/Formula/chicken.rb
+++ b/Library/Formula/chicken.rb
@@ -15,6 +15,8 @@ class Chicken < Formula
     sha1 "e2a9863f311099590265704fe410f39de802c600" => :lion
   end
 
+  depends_on 'homebrew/dupes/make' => :build
+
   def install
     ENV.deparallelize
 


mistydemeo:
Sorry I haven't merged this yet, but I'm going to test this this weekend.

mistydemeo:
> What I would really like is to be able to dep on make, then use the Cellar'd make binaries without requiring brew link.

Unfortunately not; Homebrew always requires that a dependency be linked if it's not `keg_only`. This is true of things other than make.

qu1j0t3:
I realise other tools than `make` could be build deps. I just think that the ability to use them for the duration of a build, without changing the user's configuration, would be a useful facility. Look forward to the results of your tests!

mistydemeo:
Merged with some tweaks as 4d34a375d1a8f4c5d63b056a9381c6afdaaea5b6 - thank you! I added bottles for both Tiger and Leopard, too.

PR #362, opened 2015-10-25
Removal of erroneous lines in ffmpeg Formula.

dcwar:
Lines 123, 125, and 128 in the original formula appear to be erroneously merged in. This patch removes those lines, as `brew outdated` will not function with them present.


mistydemeo:
Good catch, thank you!

dcwar:
Welcome!

PR #366, opened 2015-11-10
Make sha256 more resilient, with a platform-independent fallback.

geoff-codes:
Hi @mistydemeo!

I've been running into trouble with that binary sha2 (I'm running 10.4 Intel), and seeing in the history the tribulations you've gone through to get a decent sha256 going, I thought I'd put a bit of effort into trying to craft something a bit more robust. To sum it up, we [quickly check](https://github.com/geoff-codes/tigerbrew/blob/478902c61a4eb2ae399d6015887e6fdb59421e96/Library/Homebrew/vendor/sha256) to see if we have any of a few utilities with a fast sha256, and failing that, fall back to a pure-perl sha256 implementation which should work just about anywhere.

Let me know if you have any thoughts or questions!


geoff-codes:
Rebased...

mistydemeo:
Sorry for losing track of this.

I've since rebuilt sha256 for Intel, but the perl and openssl solutions here look great - going to look at using this or a variation on it!

mistydemeo:
This was rebased and merged as #513, with some tweaks. Thank you again!

PR #412, opened 2016-03-20
Old OSX fixes for D-Bus formula (see homebrew#50219)

zbentley:
The OnDemand key in dbus's included plist file should be uncommented
on old OSX versions. There's already a different patch for newer OSX versions in the middle of submission to the main homebrew repo; I'll keep that patch included here and also include one for pre-snow-leopard OSXes.

Other changes included:
- Removed pointless "chmod" of the distributed plist file. Unless built as root, the generated plist has appropriate permissions in standard Homebrew installs. 
- Added "master" branch info for the recommended public-pull source repo for D-Bus.


mistydemeo:
Thank you! Merged.

PR #413, opened 2016-03-22
Update D-Bus Formula

zbentley:
The D-Bus formula can be updated:
- A new version can be pulled from homebrew/homebrew. It builds on OSX 10.5 (or it did for me, at least). 
- Bottle URLs can be updated for new OSXes on the new version.
- Patch URLs from my previous patch can be updated since they are now hosted on the main patch repo (see homebrew/patches#15). 
- The git master URL can be updated to use HTTPS (see discussion on homebrew/homebrew#50219)

This formula could theoretically be used everywhere, but the mainline maintainers wanted to avoid accumulating legacy-OSX support cruft, so pre-snow-leopard installs can remain supported here.


mistydemeo:
Looks like I forgot to close this when I merged the PR.

zbentley:
Thanks!

PR #434, opened 2016-10-13
zsh: update version

awesomebing1:
I updated zsh's version from 5.1.1 to 5.2 mostly because the 5.1.1 version isn't available in those locations anymore (moved to the old versions folder) and because 5.2 is newer. I have tested this (including building + actually using zsh built).


mistydemeo:
Thank you!

PR #440, opened 2016-11-06
Updates for several tiger/homebrew formula, and tigerbrew's compilers…

dcwar:
… selection (so GCC 6 can be used with the --cc=gcc-6 switch when installing/upgrading).

I'm not an expert on this, but I'm trying. :)


mistydemeo:
Thank you!

Issue #427, opened 2016-07-01
Error: undefined method `captures' for nil:NilClass


cneira:
I'm not able to install packages using brew, brew install fails with the following message

feist:~ cnb$ brew install git
==> Installing dependencies for git: xz, gnu-tar, cctools-headers, dyld-h
==> Installing git dependency: xz
==> Downloading https://fossies.org/linux/misc/xz-5.2.1.tar.gz
Already downloaded: /Library/Caches/Homebrew/xz-5.2.1.tar.gz
Error: undefined method `captures' for nil:NilClass
Please report this bug:
    https://github.com/mistydemeo/tigerbrew/wiki/troubleshooting
/usr/local/Library/Homebrew/extend/pathname.rb:293:in`sha256_vendored_sha2'
/usr/local/Library/Homebrew/extend/pathname.rb:284:in `sha256'
/usr/local/Library/Homebrew/extend/pathname.rb:298:in`send'
/usr/local/Library/Homebrew/extend/pathname.rb:298:in `verify_checksum'
/usr/local/Library/Homebrew/resource.rb:120:in`verify_download_integrity'
/usr/local/Library/Homebrew/resource.rb:80:in `stage'
/usr/local/Library/Homebrew/formula.rb:1405:in`stage'
/usr/local/Library/Homebrew/formula.rb:859:in `brew'
/usr/local/Library/Homebrew/build.rb:106:in`install'
/usr/local/Library/Homebrew/build.rb:175
feist:~ cnb$ brew doc
Error: Unknown command: doc
feist:~ cnb$ brew doctor
Please note that these warnings are just used to help the Tigerbrew maintainers
with debugging if you file an issue. If everything you use Tigerbrew for is
working fine: please don't worry and just ignore them. Thanks!

Warning: The system curl on 10.6 and below is often incapable of supporting
modern secure connections & will fail on fetching formulae.
We recommend you:
  brew install curl

Warning: Git could not be found in your PATH.
Tigerbrew uses Git for several internal functions, and some formulae use Git
checkouts instead of stable tarballs. You may want to install Git:
  brew install git

Warning: Unbrewed dylibs were found in /usr/local/lib.
If you didn't put them there on purpose they could cause problems when
building Tigerbrew formulae, and may need to be deleted.

Unexpected dylibs:
    /usr/local/lib/libfuse.0.dylib
    /usr/local/lib/libulockmgr.1.dylib

Warning: Unbrewed header files were found in /usr/local/include.
[...]

PR #458, opened 2017-01-15
Rebuild `sha256` to fix it on Tiger i386

### LINKED ISSUE #359: curl failing on 10.4.11-i386
https://gist.github.com/picklefarmer/916dc234cfa89d7abf61

ranma42:
The new `sha256` binary is a prebuilt copy of Aaron Gifford's sha2
available at https://www.aarongifford.com/computers/sha.html

The binary is built for PowerPC and i386 using the command line:

    cc -Wall -O2 -arch i386 -arch ppc -mmacosx-version-min=10.4 \
    -isysroot/Developer/SDKs/MacOSX10.4u.sdk \
    -DMACOSX_DEPLOYMENT_TARGET=10.4 sha2.c sha2prog.c -o sha2prog

Fixes #359 and #427.

This is an alternative to #366.


ranma42:
Is there anything I should do to make this progress?

I realize that a pre-compiled binary generated from an external developer might not be acceptable for inclusion in the repository. If this is a blocker, I might be able to guide you in setting up a 10.4 VM to rebuild the binary on your machine.

Alternatively, would it be feasible for you to try and regenerate the binary so that the _read$UNIX2003 symbol is not used? I would expect it to be possible with the appropriate compiler/linker flags.

mistydemeo:
> If this is a blocker, I might be able to guide you in setting up a 10.4 VM to rebuild the binary on your machine.

mistydemeo:
> If this is a blocker, I might be able to guide you in setting up a 10.4 VM to rebuild the binary on your machine.

This would be really helpful! I'd prefer to build this binary myself, but I haven't been successful at getting a 10.4 VM running in Virtualbox yet.

ranma42:
I tried to simplify the procedure as much as possible and document it [here](https://github.com/ranma42/TigerOnVBox).
Sorry if it took some time, but I wanted to try some different options.

On the bright side, I now have a much better environment than before, as it is running on the SATA controller instead of the IDE (with significantly improved performance) and I got rid of an annoying issue of two cursors being shown (one by the guest OS, one by the host OS) in slightly different positions.

Please let me know if the procedure works for you.

mistydemeo:
Thanks so much for this - it works a charm! Got it working, and rebuilt sha2 in #465. It's working for me, can you test to see if it works for you?

PR #460, opened 2017-01-19
cairo: fix build on 10.4

### LINKED ISSUE #452: cairo will not build on Mac OS 10.4
When building cairo 1.14.2 like this: 'brew -v install cairo', the build fails with this error message:

cairo-quartz-font.c: In function ‘_cairo_quartz_init_glyph_path’:
cairo-quartz-font.c:548: error: ‘CTFontRef’ undeclared (first use in this function)
cairo-quartz-font.c:548: error: (Each undeclared identifier is reported only once
cairo-quartz-font.c:548: error: for each function it appears in.)
cairo-quartz-font.c:548: error: parse error before ‘ctFont’
cairo-quartz-font.c:563: error: ‘ctFont’ undeclared (first use in this function)
cairo-quartz-font.c:563: error: implicit declaration of function ‘CTFontCreateWithGraphicsFont’
cairo-quartz-font.c:564: error: implicit declaration of function ‘CTFontCreatePathForGlyph’
make[3]: *** [Makefile:2342: cairo-quartz-font.lo] Error 1

mistydemeo:
Also updates to the latest version.

Fixes #452.

cc @ranma42 - mind testing this?

### DIFF
--- a/Library/Formula/cairo.rb
+++ b/Library/Formula/cairo.rb
@@ -1,18 +1,9 @@
 class Cairo < Formula
   desc "Vector graphics library with cross-device output support"
   homepage "http://cairographics.org/"
-  url "http://cairographics.org/releases/cairo-1.14.2.tar.xz"
-  mirror "https://www.mirrorservice.org/sites/ftp.netbsd.org/pub/pkgsrc/distfiles/cairo-1.14.2.tar.xz"
-  sha256 "c919d999ddb1bbbecd4bbe65299ca2abd2079c7e13d224577895afa7005ecceb"
-  revision 1
-
-  bottle do
-    revision 3
-    sha256 "c53e4efa75921181a8689a0dcd13454d7a0759e259b331ebdc92c692d0716bad" => :el_capitan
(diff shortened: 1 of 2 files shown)


ranma42:
`brew install cairo` in my environment fails because glib seems to require a newer python (this [gist](https://gist.github.com/ranma42/c2bc7a648f47fd6fe32fc5cff0eaabf3) shows my env and the error).

`brew install python` seems to fix the installation of the glib dependency, which then makes cairo install just fine as shown [here](https://gist.github.com/ranma42/7a073e847cceb99af7c87da9c18e27ba)

ranma42:
The `configure.ac` file for glib seems to indicate that it requires [python 2.5 or newer](https://github.com/GNOME/glib/blob/40cf271a779ed2bb29361b6e9fb6b9f814204d47/configure.ac#L395). MacOSX 10.4 (even if you install XCode 2.5) ships with python 2.3.5, so maybe the right fix is to add `python` as a dependency for `glib` (tigerbrew already handles glib on 10.4 in a special way because of the outdated zlib).

mistydemeo:
Added the Python dep.

PR #466, opened 2017-04-02
cairo: Fix building Cairo w/ X11 on Tiger

### LINKED ISSUE #69: Cairo can't build with Tiger's X11
Tiger's X11 is far, far too old to build cairo. Cairo mostly is able to tell that this is the case, but still tries to build xlib-xrender and the build consequently fails. Build logs:

https://gist.github.com/anonymous/5695285

X has been disabled as of 4e19c7f94873ae955a5682c6d2d8e41a2645f503. Users who need X on Cairo, let me know and we can see if we can fix it up.

ranma42:
Tiger ships with an old X11 that does not include the XCB (X protocol
C-language Binding), but the Xlib backend builds just fine.

The build failure at https://gist.github.com/anonymous/5695285 is
caused by a defect in the cairo build system that has been fixed
upstream.

Fixes #69.

### DIFF
--- a/Library/Formula/cairo.rb
+++ b/Library/Formula/cairo.rb
@@ -10,7 +10,7 @@ class Cairo < Formula
   option :universal
 
   depends_on "pkg-config" => :build
-  depends_on :x11 => :recommended if MacOS.version == :leopard
+  depends_on :x11 => :recommended if MacOS.version <= :leopard
   depends_on :x11 => :optional if MacOS.version > :leopard
   depends_on "freetype"
   depends_on "fontconfig"
@@ -40,7 +40,9 @@ class Cairo < Formula
     ]
 
     if build.with? "x11"
-      args << "--enable-xcb=yes" << "--enable-xlib=yes" << "--enable-xlib-xrender=yes"
+      # Tiger's X11 does not include XCB
+      args << "--enable-xcb=yes" if MacOS.version > :tiger
+      args << "--enable-xlib=yes" << "--enable-xlib-xrender=yes"
     else
       args << "--enable-xcb=no" << "--enable-xlib=no" << "--enable-xlib-xrender=no"
     end
--- a/Library/Formula/cairo.rb
+++ b/Library/Formula/cairo.rb
@@ -41,7 +41,7 @@ class Cairo < Formula
 
     if build.with? "x11"
       # Tiger's X11 does not include XCB
-      args << "--enable-xcb=yes" if MacOS.version > :tiger
+      args << "--enable-xcb=yes" if MacOS.version > :leopard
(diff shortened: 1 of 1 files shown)


ranma42:
WARNING: I only tried it on Tiger.

I would like to test the build on Leopard, but I do not have access to a 10.5 build environment at this time. Specifically, I would like to check if the cairo-xcb backend (and xcb-shm) is enabled on Leopard.

mistydemeo:
I'll test this on a Leopard VM tomorrow.

Library/Formula/cairo.rb
```diff
@@ -8,11 +8,9 @@ class Cairo < Formula
   keg_only :provided_pre_mountain_lion
 
   option :universal
-  # Tiger's X11 is simply way too old
-  option 'without-x', 'Build without X11 support' if MacOS.version > :tiger
 
   depends_on "pkg-config" => :build
-  depends_on :x11 => :optional if MacOS.version > :leopard
+  depends_on :x11 => if MacOS.version > :leopard then :optional else :recommended end
```
mistydemeo: I'd prefer to see this put into multiple lines for readability:

```ruby
if MacOS.version > :leopard
  depends_on :x11 => :optional
else
  depends_on :x11 => :recommended
end
```

Incidentally, why is this recommended on Tiger?

ranma42: I removed the `option 'without-x'` and I made it `:recommended` to preserve the old behaviour.
Then I extended it from just `:leopard` to both `:tiger` and `:leopard`.
I (rebased and) pushed a new patchset that should make this steps more visible.

ranma42:
Oops, [my reply](https://github.com/mistydemeo/tigerbrew/pull/466#discussion_r109305637) is very well hidden after the rebase.

Library/Formula/cairo.rb
mistydemeo: Thanks!

mistydemeo:
Looks like it doesn't work due to an outdated XCB:

> checking whether cairo's XCB surface backend feature could be enabled... no (requires xcb >= 1.6 xcb-render >= 1.6 http://xcb.freedesktop.org)

At some point I'm going to look at backporting a newer X11, but that day is not today. 😅

mistydemeo:
Specifically, here's what config.log says:

```
configure:22769: $PKG_CONFIG --exists --print-errors "$xcb_REQUIRES"
Requested 'xcb >= 1.6' but version of XCB is 1.1
Requested 'xcb-render >= 1.6' but version of XCB Render is 1.1
```

mistydemeo:
Merged with a tweak to fix Leopard - thanks again!

PR #470, opened 2017-04-04
Fix building GCC on 10.4 Intel

mistydemeo:
This is still a WIP; GCC still fails to build with these patches at this time.


mistydemeo:
Confirmed working with the last commit.

PR #475, opened 2017-04-04
ruby: fix build on Tiger

### LINKED ISSUE #473: Ruby 2.2.3 will not build on i386 Tiger
https://gist.github.com/14fa08329609ac8132538c054bec33f7

Relevant part, from signal.c:

```
compiling signal.c
signal.c: In function ‘check_stack_overflow’:
signal.c:778: error: ‘struct mcontext’ has no member named ‘__ss’
make: *** [signal.o] Error 1
```

The code in question:

```c
# elif defined __APPLE__
#   if defined(__LP64__)
    const uintptr_t sp = mctx->__ss.__rsp;
#   else
    const uintptr_t sp = mctx->__ss.__esp;
#   endif
```

mistydemeo:
The inreplace is obviously an ugly, ugly hack; I'm just fitting this in for now until I report this upstream and there's a real patch.

Fixes #473.

### DIFF
--- a/Library/Formula/ruby.rb
+++ b/Library/Formula/ruby.rb
@@ -35,6 +35,27 @@ class Ruby < Formula
   end
 
   def install
+    # mcontext types had a member named `ss` instead of `__ss`
+    # prior to Leopard; see
+    # https://github.com/mistydemeo/tigerbrew/issues/473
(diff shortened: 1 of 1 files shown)


mistydemeo:
https://bugs.ruby-lang.org/issues/13400

PR #476, opened 2017-04-04
ruby 2.4.1

mistydemeo:
Build currently failing due to use of `fgetattrlist`, first introduced in 10.6.


ranma42:
2.4.1 seem to require several patches in order to build:
 - https://svn.ruby-lang.org/cgi-bin/viewvc.cgi/trunk/dir.c?r1=56608&r2=57248
 - https://svn.ruby-lang.org/cgi-bin/viewvc.cgi/trunk/ext/-test-/memory_status/memory_status.c?r1=56377&r2=57180

With these patches applied, the build still fails for me on `ext/fiddle`

PR #477, opened 2017-04-04
ruby: Fix build of 2.4.1 on Tiger i386

ranma42:
Even though I did not test it, I think that the first two patches are needed on Leopard as well, because:
 - `fgetattrlist` was introduced in 10.6
 - `mach_task_basic_info_data_t` is not available even on 10.6

I am not sure about `win32.S` (maybe not needed on Leopard? likely not needed on ppc).

### DIFF
--- a/Library/Formula/ruby.rb
+++ b/Library/Formula/ruby.rb
@@ -27,6 +27,24 @@ class Ruby < Formula
   depends_on "openssl"
   depends_on :x11 if build.with? "tcltk"
 
+  if MacOS.version <= :leopard
+    # fix for https://bugs.ruby-lang.org/issues/11054
+    patch do
+      url "https://github.com/ruby/ruby/commit/1c80c388d5bd48018c419a2ea3ed9f7b7514dfa3.patch"
+      sha256 "b53661f3077f10af2446f6faf5870d476471290859b7058cd2661e6e6ca24e8e"
+    end
+
+    # fix for https://bugs.ruby-lang.org/issues/13247
+    patch do
+      url "https://github.com/ruby/ruby/commit/9e1a9858c84142e32b1bc51b23fa06a025f98b46.patch"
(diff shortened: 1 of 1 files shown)


mistydemeo:
Thank you!

ranma42:
While fiddling with OpenSSL I found out that the assembler provided by `cctools` can parse `.balign` pseudo-ops. This means that adding `cctools` as build dependency on Tiger would be an alternative to patching `ext/fiddle/libffi-3.2.1/src/x86/win32.S` (which in turn would mean that ruby 2.4.2 would likely require no patching at all).
Basically the same applies to icu4c. If this approach looks better to you as well, I can do a PR to conditionally depend on `cctools` instead of patching the assembly files for each of these formulas.

I was wondering if it would make sense for the Tiger and Leopard build environment to always require `cctools` and `ld64` unless the formula explicitly opts out. This would likely make the two build environments more similar and possibly easier to manage.

mistydemeo:
> I was wondering if it would make sense for the Tiger and Leopard build environment to always require cctools and ld64 unless the formula explicitly opts out. This would likely make the two build environments more similar and possibly easier to manage.

Longterm, I think that makes sense. I'd like to see that rolled into the larger series of changes that starts providing binary packages for common dependencies. Over the next few months, I'll be rolling out a series of changes that causes Tigerbrew to depend on binary curl and Ruby packages, downloaded at startup if they're not available. Distributing statically-linked ld64 and cctools packages the same way, and having those unconditionally available and used by default, makes sense to me.

At the same time, I'll be splitting Tigerbrew into separate core/formula repos like Homebrew, and migrating Tigerbrew to use the same core code as Homebrew. Given the overarching similar concerns to Homebrew, it makes sense to me to have Tigerbrew be a unique core tap that uses the same application code, rather than a 100% separate codebase. Most of the core code differences between Tigerbrew and Homebrew have already been merged into Homebrew at this point, in preparation for that.

ranma42:
I would love this! As you have probably noticed, I am still learning the internals of (home)brew, but I am more than willing to try and help :)

Issue #70, opened 2013-06-03
Harfbuzz won't link against icu4c


mistydemeo:
https://gist.github.com/5695545

icu4c's install names use a token that Tiger's ld can't understand. Even when trying to force ld64, however, harfbuzz builds with the system ld and consequently fails on the icu4c step.

icu4c is currently disabled in harfbuzz on Tiger, but a better solution is needed.

PR #481, opened 2017-04-09
Fix harfbuzz

ranma42:
This is sufficient to fix the build for harfbuzz in my VM.
Please confirm whether it works in other environments, as issue #70 mentions that the build is not fixed by `ld64`.


mistydemeo:
Hm, still failing for me, due to a gobject-introspection issue: https://gist.github.com/3827abf6034629c8cf10a5e24aa2d83c

ranma42:
Sorry, you're right, the default `make` command cannot handle the makefiles from gobject-introspection. Added the dependency to `[g]make`

PR #488, opened 2017-04-17
Update OpenSSL and enable universal build on Tiger

ranma42:
The main aim of this PR is updating the OpenSSL formula and fixing it to build a universal library on Tiger (with assembly-optimized functions).

In order to do so, some additional changes are applied to `Library/Homebrew/os/mac/hardware.rb`.
Specifically, the detection of 64-bit CPUs now works on Tiger and the architectures returned by `universal_archs` now depend on the os, but not on the CPU.

### DIFF
--- a/Library/Homebrew/os/mac/hardware.rb
+++ b/Library/Homebrew/os/mac/hardware.rb
@@ -79,7 +79,11 @@ module MacCPUs
   end
 
   def bits
-    sysctl_bool("hw.cpu64bit_capable") ? 64 : 32
+    # hw.cpu64bit_capable is not available on Tiger, but it is
+    # possible to detect 64-bit CPUs by checking the ppc- and
+    # intel-specific flags, as shown in
+    # https://opensource.apple.com/source/xnu/xnu-1504.7.4/tools/tests/xnu_quick_test/misc.c?txt
+    sysctl_bool("hw.optional.64bitops") || sysctl_bool("hw.optional.x86_64") ? 64 : 32
   end
 
   def arch_32_bit
--- a/Library/Homebrew/os/mac/hardware.rb
+++ b/Library/Homebrew/os/mac/hardware.rb
@@ -97,12 +97,15 @@ module MacCPUs
   # Returns an array that's been extended with ArchitectureListExtension,
   # which provides helpers like #as_arch_flags and #as_cmake_arch_flags.
   def universal_archs
-    # Building 64-bit is a no-go on Tiger, and pretty hit or miss on Leopard.
-    # Don't even try unless Tigerbrew's experimental 64-bit Leopard support is enabled.
-    if MacOS.version <= :leopard && !MacOS.prefer_64_bit?
-      [arch_32_bit].extend ArchitectureListExtension
+    # Support for running on ppc was dropped after Leopard
+    if MacOS.version > :leopard
+      [:x86_64, :i386].extend ArchitectureListExtension
+    # Building 64-bit is pretty hit or miss on Tiger and Leopard.
+    # Don't even try unless Tigerbrew's experimental 64-bit support is enabled.
+    elsif ENV["HOMEBREW_PREFER_64_BIT"]
+      [:x86_64, :i386, :ppc64, :ppc].extend ArchitectureListExtension
     else
-      [arch_32_bit, arch_64_bit].extend ArchitectureListExtension
+      [:i386, :ppc].extend ArchitectureListExtension
     end
   end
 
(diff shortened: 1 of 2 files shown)


Library/Homebrew/os/mac/hardware.rb
```diff
@@ -79,7 +79,11 @@ def cores
   end
 
   def bits
-    sysctl_bool("hw.cpu64bit_capable") ? 64 : 32
+    # hw.cpu64bit_capable is not available on Tiger, but it is
+    # possible to detect 64-bit CPUs by checking the ppc- and
+    # intel-specific flags, as shown in
+    # https://opensource.apple.com/source/xnu/xnu-1504.7.4/tools/tests/xnu_quick_test/misc.c?txt
+    sysctl_bool("hw.optional.64bitops") || sysctl_bool("hw.optional.x86_64") ? 64 : 32
```
mistydemeo: Does this work on more recent versions of OS X, too?

Library/Homebrew/os/mac/hardware.rb
```diff
@@ -93,12 +97,15 @@ def arch_64_bit
...
-    # Building 64-bit is a no-go on Tiger, and pretty hit or miss on Leopard.
-    # Don't even try unless Tigerbrew's experimental 64-bit Leopard support is enabled.
-    if MacOS.version <= :leopard && !MacOS.prefer_64_bit?
-      [arch_32_bit].extend ArchitectureListExtension
+    # Support for running on ppc was dropped after Leopard
+    if MacOS.version > :leopard
+      [:x86_64, :i386].extend ArchitectureListExtension
+    # Building 64-bit is pretty hit or miss on Tiger and Leopard.
+    # Don't even try unless Tigerbrew's experimental 64-bit support is enabled.
+    elsif ENV["HOMEBREW_PREFER_64_BIT"]
```
mistydemeo: I'd still like to see this guarded by the `MacOS.prefer_64_bit?` check, instead of the environment variable; `prefer_64_bit?` checks the env var, and I don't want the rest of the app to be aware of it.

Library/Homebrew/os/mac/hardware.rb
ranma42: Yes, `hw.optional.x86_64` is available on all recent versions (of course `hw.optional.64bitops` will be absent, as it is only available when the architecture is ppc). It is also kind of documented in the [`sys/sysctl.h` header](https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.11.sdk/usr/include/sys/sysctl.h#L669-L687).

Library/Homebrew/os/mac/hardware.rb
ranma42: At first I did that, but then I found out that the result of `MacOS.prefer_64_bit?` depends on whether the CPU is 64 bits or not. Instead, I wanted to change `universal_archs` so that it only depends on the target operating system (+ environment flags), because a 32-bit system is perfectly capable of building a 64-bit library.

On a side note, it would be interesting to extend `MacOS.prefer_64_bit?` to also allow for building 64-bit binaries on Tiger; I would expect it to be quite unreliable, just like on Leopard, essentially because several of the system libraries only come as i386/ppc, and very few of them are i386/ppc.

mistydemeo: That's a good point. I'll add a new helper to encapsulate this, something like `MacOS.allow_64_bit`

mistydemeo:
Encountered an issue on PowerPC hosts. The libraries in `/usr/lib` contain only PPC slices, unlike on Intel hosts where the libraries in `/usr/lib` contain all four slices. Right now the build will prefer linking against the `/usr/lib` copies of libraries, instead of the 10.4u SDK copies, so it'll fail while linking.

ranma42:
What is the best way to fix this? Should the build use the 10.4u SDK or should I restrict the `universal_archs` list to `:ppc, :ppc64` on ppc?

mistydemeo:
I think some adjustments are needed on the Homebrew/Tigerbrew side to adjust the build flags so that `SDKROOT` is set to the SDK path even when the CLT is installed (as it always is in Xcode 2.x and 3.x). That might have knock-on effects though. :|

lyssdod:
@mistydemeo @ranma42 Hi, can you clarify the status of this? Just tried to install openssl on Tiger and it took 17 minutes for my poor G4 to compile it...and the stuff built is even without assembly speedups.

Also, why there's no binary version distributed? I've seen some packages use us.archive.org

Thanks

PR #510, opened 2017-05-11
Add support for installing a vendored Ruby

mistydemeo:
This is based mostly on the upstream vendor-ruby code, using an i386/ppc Ruby I built.

TODOs:

- [x] Add a fallback sha256 tool to replace the C `sha2`. The chicken-and-egg of calculating portable-openssl's checksum still needs to be solved, but I no longer want to be shipping a prebuilt `sha2` binary in the source tree. I'll take a look at @geoff-codes's work in #366.
- [x] Fix `flock` code; I have a PR for this upstream at https://github.com/Homebrew/brew/pull/2624 that needs more attention.


mistydemeo:
Added vendored openssl and curl, too. This is pretty close to shippable, and will fix a lot of problems. Two remaining things to do, which I've added to the PR summary.

ranma42:
Modern Ruby versions should provide both sha256 and flock.
Would it be possible to use the Perl implementations only for the bootstrap and then use the internal Ruby implementation? This would probably avoid some Tiger-specific code paths in the main codebase (they would still be present in the bootstrap/install-vendor steps)

mistydemeo:
> Would it be possible to use the Perl implementations only for the bootstrap and then use the internal Ruby implementation?

Yup, planning on using either the Perl or Python version. Unfortunately, nothing shipped on Tiger supports sha256 - the modern openssl is needed to do that.

mistydemeo:
`flock` fix merged upstream.

EricFromCanada:
Looking forward to this. Currently this branch fails on my 10.5 Intel VM with "Failed to install vendor ruby". When done, will this allow commands like brew's `test-bot`, `test`, `tests`, `audit` to start working?

Library/Homebrew/cmd/vendor-install.sh
```diff
@@ -0,0 +1,220 @@
...
+
+VENDOR_DIR="$HOMEBREW_LIBRARY/Homebrew/vendor"
+
+# Built from https://github.com/Homebrew/homebrew-portable.
+if [[ -n "$HOMEBREW_OSX" ]]
+then
+  # Universal 10.4 build
+  if [[ "$HOMEBREW_OSX_VERSION_NUMERIC" -lt "100500" || "$HOMEBREW_PROCESSOR" != "Intel" ]]
+  then
+    ruby_URL="https://archive.org/download/tigerbrew/portable-ruby-2.2.7.tiger.bottle.tar.gz"
```
MikeMcQuaid: May be worth going for 2.3.3p222 here to match High Sierra.

mistydemeo:
I've got this set up with a confirmed-working-on-G3 Ruby and Curl, along with i386 builds of both packages as well.

@EricFromCanada Mind giving this a second go? I think I'm ready to ship this.

EricFromCanada:
On G3/10.4 & Intel/10.5, after running `git checkout vendor_ruby` in `/usr/local` and then any `brew` command, it populates `/usr/local/Library/Homebrew/vendor` as expected. Would be nice if there was some output while it's pulling down the portables, though.

✅ `brew audit git`
✅ `brew test git`
✅ `brew install xtitle`
⛔️ `brew doctor`
```
Warning: You have uncommitted modifications to Tigerbrew
If this a surprise to you, then you should stash these modifications.
Stashing returns Tigerbrew to a pristine state but can be undone
should you later need to do so for some reason.
    cd /usr/local/Library && git stash && git clean -d -f

Warning: Ruby version 2.3.3 is unsupported on 10.5. Tigerbrew
is developed and tested on Ruby 1.8, and may not work correctly
on other Rubies. Patches are accepted as long as they don't cause breakage
on supported Rubies.
$ git status
On branch vendor_ruby
Your branch is up-to-date with 'origin/vendor_ruby'.
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	Library/Homebrew/vendor/portable-curl/
	Library/Homebrew/vendor/portable-ruby/

nothing added to commit but untracked files present (use "git add" to track)
```
⛔️ `brew tests`
```
ERROR:  While executing gem ... (OptionParser::InvalidOption)
    invalid option: --user-install
Error: Failure while executing: gem install --no-ri --no-rdoc --user-install bundler
```
although running it manually worked, sort of:
```
$ /usr/local/Library/Homebrew/vendor/portable-ruby/current/bin/gem install --no-ri --no-rdoc --user-install bundler
WARNING:  You don't have /Users/serveradmin/.gem/ruby/2.3.0/bin in your PATH,
	  gem executables will not run.
Successfully installed bundler-1.16.1
1 gem installed
```
One more note: `brew install --verbose <formula>` shows `curl` still sending a user agent string with the old version number, i.e.
```
/usr/local/Library/Homebrew/vendor/portable-curl/current/bin/curl -fLA Homebrew/0.9.5 (Macintosh; powerpc Mac OS X 10.4.11) curl/7.16.3 …
```

mistydemeo:
> One more note: brew install --verbose <formula> shows curl still sending a user agent string with the old version number, i.e.

Oops - this is a bug. That value is set *before* the vendored curl is set up, which means the user agent references a different curl from the one that's used. I'll fix that!

> brew tests

I'll look into this.

> brew doctor

Silly bug, but easy fix. 😅 I'll remove the bad test.

mistydemeo:
> brew tests

Aha - this is because the portable Ruby doesn't include Bundler. I'm not sure which `bundle` it ended up running.

PR #525, opened 2017-07-13
Revert to QEMU 2.2.1

### LINKED ISSUE #523: QEMU not working/freezing
Hi, I've installed QEMU on my 2x2.0 GHz G5 machine and it almost always freezes at the SeaBIOS (version rel-1.8.2-0-g33fbe13 by qemu-project.org screen). I'm wondering if this is due to some saying that versions past 2.2.0 don't work on pre-10.6 systems due to this:
"Okay, okay, this is the PowerPC equivalent of a stupid pet trick, but this is qemu 2.2.1 (the last version that did not require thread-local storage, which won't work with OS X 10.6 and earlier)"
(Taken from a page written by the developer of TenFourFox: http://tenfourfox.blogspot.com/2017/02/4580b1-available-with-user-agent.html)

So most people install 2.2.0 through MacPorts 2.3.4, should that be what's happening here or is there just something wrong with my install? Does 2.4.0.1 work on other machines that installed it using tigerbrew?

I've tried a lot of different launch options, but no matter what variation I use the same issue happens.

mistydemeo:
Fixes #523.

cc @LarBob - I'm still testing this, but it probably doesn't work yet.

### DIFF
--- a/Library/Formula/qemu.rb
+++ b/Library/Formula/qemu.rb
@@ -39,10 +39,16 @@ class Qemu < Formula
       --prefix=#{prefix}
       --cc=#{ENV.cc}
       --host-cc=#{ENV.cc}
-      --enable-cocoa
       --disable-bsd-user
       --disable-guest-agent
     ]
+
+    # Cocoa UI uses features that require 10.5 or newer
+    if MacOS.version > :tiger
+      args << "--enable-cocoa"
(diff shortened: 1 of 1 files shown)


mistydemeo:
@larbob Mind seeing if this fixes it for you? Builds now.

PR #530, opened 2017-09-11
curl: add HTTP mirror.

### LINKED ISSUE #506: brew install curl fails
I mentioned it in <https://github.com/mistydemeo/tigerbrew/issues/500>: after initial `ruby -e "$(curl -fsSkL raw.github.com/mistydemeo/tigerbrew/go/install)"` and `brew doctor`, I try to get new curl: `brew install curl`. This started failing yesterday evening and still not working today. I tested on PPC Tiger Server on Mac Mini G4, PPC Leopard on iMac G5 and on Intel Tiger VM. It seems curl bundled with Mac OS X is unable to download from `https://curl.haxx.se/download/curl-7.50.3.tar.bz2`. I just checked and I can download this file fine from modern browser.

I think it would work fine if you can switch to `http://curl.haxx.se/download/curl-7.50.3.tar.bz2`. You still have sha256 checksum to verify integrity.

Alternatively, please consider creating a simple boilerplate site `http://tigerbrew.github.io` and host mirrors of critical tarballs and may be bottled versions of key packages like curl, git, apple-gcc42, lld64 etc.

MikeMcQuaid:
This allows a curl that cannot handle modern SSL certificates to install a newer version.

Fixes #506

### DIFF
--- a/Library/Formula/curl.rb
+++ b/Library/Formula/curl.rb
@@ -2,6 +2,7 @@ class Curl < Formula
   desc "Get a file from an HTTP, HTTPS or FTP server"
   homepage "https://curl.haxx.se/"
   url "https://curl.haxx.se/download/curl-7.50.3.tar.bz2"
+  mirror "http://curl.askapache.com/download/curl-7.50.3.tar.bz2"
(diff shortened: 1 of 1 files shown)


djdrzzy:
This works awesome. Thanks!

Issue #32, opened 2013-05-09
gmp-5.11 will not build (Leopard)


ghost:
tmp-add_n.s:98:rldicl instruction is only for 64-bit implementations (not allowed without -force_cpusubtype_ALL option)

same file, same issue 'std', line 100.

PR #549, opened 2018-01-09
Upgrade GCC to 7.2.0

mistydemeo:
cc @mmolinac - I'm currently testing this as a potential fix. If this doesn't work, I'm going to attempt passing the `-force_cpusubtype_ALL` flag; this fixed a similar issue in #32.

### DIFF
--- a/Library/Formula/gcc.rb
+++ b/Library/Formula/gcc.rb
@@ -21,15 +21,9 @@ class Gcc < Formula
 
   desc "GNU compiler collection"
   homepage "https://gcc.gnu.org"
-  url "http://ftpmirror.gnu.org/gcc/gcc-5.2.0/gcc-5.2.0.tar.bz2"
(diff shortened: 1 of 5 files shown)


mistydemeo:
Added support for building with GCC 7, and tested it:

```
$ brew install sha2 --cc=gcc-7
==> Downloading https://www.aarongifford.com/computers/sha2-1.0.1.tgz
Already downloaded: /Library/Caches/Homebrew/sha2-1.0.1.tgz
==> /usr/local/bin/gcc-7 -o sha2 sha2prog.c sha2.c
==> perl sha2test.pl
/usr/local/Cellar/sha2/1.0.1: 3 files, 52K, built in 9 seconds
```

In addition, I ran all of the test programs defined in the GCC formula. Everything is working. Think we're good here. Thanks for your help with this, @mmolinac!

PR #569, opened 2018-02-25
Refuse to try git installs on old OS Xs due to TLS

mistydemeo:
cc @cjbayliss - this should fix the error you encountered.

### DIFF
--- a/install
+++ b/install
@@ -4,6 +4,12 @@
 # change the value of HOMEBREW_PREFIX.
 HOMEBREW_PREFIX = '/usr/local'
 
+CURL_VERSION    = "7.58.0"
+CURL_PPC_URL    = "https://archive.org/download/tigerbrew/portable-curl-#{CURL_VERSION}.tiger_g3.bottle.tar.gz"
+CURL_PPC_SHA1   = "5bbd587871bca54637437359555618a4b1a612ec"
+CURL_INTEL_URL  = "https://archive.org/download/tigerbrew/portable-curl-#{CURL_VERSION}.tiger_i386.bottle.tar.gz"
+CURL_INTEL_SHA1 = "6e6e7989adb20856c0473aedecd623f443a6ac55"
+
 module Tty extend self
   def blue; bold 34; end
   def white; bold 39; end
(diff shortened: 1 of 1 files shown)


cjbayliss:
Running that results in this error:

```
[cjb@fedora ~]$ ruby install
==> This script will install:
/usr/local/bin/brew
/usr/local/Library/...
/usr/local/share/man/man1/brew.1

Press ENTER to continue or any other key to abort
==> Downloading and Installing Homebrew...
install:76:in `<': comparison of Float with String failed (ArgumentError)
	from install:76:in `git'
	from install:156
	from install:155:in `chdir'
	from install:155
[cjb@fedora ~]$
```

I don't know Ruby at all. But I doubt 10.11 would be considered greater than 10.7 (a mistake I've made so many times btw xD) so I changed the line to:

```ruby
return if macos_version < 10.9
```
giving us this:

```
[cjb@fedora ~]$ ruby install
==> This script will install:
/usr/local/bin/brew
/usr/local/Library/...
/usr/local/share/man/man1/brew.1

Press ENTER to continue or any other key to abort
==> Downloading and Installing Homebrew...
curl: (35) error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version
==> Installation successful!
You should run `brew doctor' *before* you install anything.
Now type: brew help
[cjb@fedora ~]$ brew doctor
-bash: brew: command not found
[cjb@fedora ~]$ ls /usr/local/Homebrew/
[cjb@fedora ~]$ ls /usr/local/bin
[cjb@fedora ~]$
```

AFAICT I'd have to use a newer ssl libary.

mistydemeo:
Good catch! Updated the git version as you suggested, and also passed `-k` to curl - this was already done on 10.5 and below, but clearly we need to do that on much newer OS X versions now.

mistydemeo:
You know, I've got an idea. :) I'll have the installer stage a copy of the statically-linked curl, and use that to download the tarball. Might be slightly overkill, but it'll do the trick. Give me an hour or two and I'll push a new commit to test.

mistydemeo:
> Also the "10.9" needs to be 10.9 so that it's a float.

I've been writing too much bash.

cjbayliss:
Ah yes, to be fair, bash is beautiful in it's own way:

```bash
while [ "0" -le 10 ]; do echo "true"; done
```

Maybe we should write a new universal language that can be exported in all the languages! /s

https://xkcd.com/927/

mistydemeo:
Can you try this latest version?

cjbayliss:
This worked after changing line 112 from:

```ruby
our_hash = `openssl sha1 curl.tar.gz`.chomp.split(" ").last
```

to:

```ruby
our_hash = `openssl sha1 /tmp/curl.tar.gz`.chomp.split(" ").last
```

And this is the output of the install process:

```
[cjb@fedora ~]$ ruby install 
==> This script will install:
/usr/local/bin/brew
/usr/local/Library/...
/usr/local/share/man/man1/brew.1

Press ENTER to continue or any other key to abort
==> Downloading and Installing Homebrew...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1836k  100 1836k    0     0   219k      0  0:00:08  0:00:08 --:--:--  333k
==> Installation successful!
You should run `brew doctor' *before* you install anything.
Now type: brew help
[cjb@fedora ~]$ brew doctor
Your system is ready to brew.
[cjb@fedora ~]$
```

Thank you so much for working on this so fast! :)

mistydemeo:
Good catch - I was testing from within `/tmp`, so that slipped past me.

Thank you for testing so quickly! I'm going to ship this. :)

PR #571, opened 2018-03-12
gcc5: update to 5.5.0, fix build on Intel Tiger

EricFromCanada:


mistydemeo:
I don't have an Intel Mac to test on, but looks good to me!

PR #572, opened 2018-03-18
gcc6: update to 6.4.0, fix build on Intel Tiger

EricFromCanada:


mistydemeo:
I have a build running at the moment. It's been chewing away on the symbol comparison step for a suspiciously long time, but the CPU is actually working and it hasn't failed yet.

mistydemeo:
It looks like the `compare-debug` phase that's hanging/taking forever is one that @sevan removes from older OSs in #954. I'm thinking I'll look at combining both of your branches and see if that gets a more successful build.

PR #580, opened 2018-04-28
mgba 0.6.3

mistydemeo:
cc @endrift

### DIFF
--- a/Library/Formula/mgba.rb
+++ b/Library/Formula/mgba.rb
@@ -1,0 +1,41 @@
+class Mgba < Formula
+  desc "Game Boy Advance emulator"
+  homepage "https://mgba.io/"
+  url "https://github.com/mgba-emu/mgba/archive/0.6.3.tar.gz"
+  sha256 "35c8906c01206adae714fee9c2bc39698c514a4adb32c9c4a5a3fa58b2d10f9a"
+  head "https://github.com/mgba-emu/mgba.git"
+
+  depends_on :ld64 # fixes a linktime error on Tiger
+  depends_on "cmake" => :build
+  depends_on "pkg-config" => :build
+  depends_on "libepoxy"
+  depends_on "libpng"
+  depends_on "libzip"
+  depends_on "sdl2"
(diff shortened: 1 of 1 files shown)


Library/Formula/mgba.rb
```diff
@@ -0,0 +1,38 @@
...
+  head "https://github.com/mgba-emu/mgba.git"
+
+  depends_on :ld64 # fixes a linktime error on Tiger
+  depends_on "cmake" => :build
+  depends_on "pkg-config" => :build
+  depends_on "libepoxy"
+  depends_on "libpng"
+  depends_on "libzip"
+  depends_on "sdl2"
+  depends_on "sqlite"
```
endrift: The sqlite-using feature isn't exposed in the SDL version. Might be better to just disable it to reduce build times.

Library/Formula/mgba.rb
```diff
@@ -0,0 +1,38 @@
...
+  fails_with :gcc if MacOS.version == :tiger
+
+  def install
+    # Qt is still broken on 10.4, so build with it off by default.
+    # FFmpeg introduces a VideoDecodeAcceleration framework requirement,
+    # which isn't available on older OS Xs.
+    # Imagemagick fails to link due to some strange path lookup issues.
+    args = ["-DBUILD_QT=OFF", "-DUSE_FFMPEG=OFF", "-DUSE_MAGICK=OFF"]
+    # GCC 4.0 and 4.2 don't support -flto
+    if [:gcc, :gcc_4_0].include? ENV.compiler
```
endrift: This is supposed to be handled by CMake. Is it not?

mistydemeo: Thanks! I'll leave a comment to remind myself to do that in the next release.

mistydemeo: I took a look at the test in master, and that won't cover every case - GCC builds without LTO on Tiger. I'll look at submitting a patch that makes the detection a bit more robust!

PR #591, opened 2018-05-24
sdl2: update SDL2 from 2.0.3 to 2.0.8

miniupnp:
Signed-off-by: Thomas Bernard <miniupnp@free.fr>

### DIFF
--- a/Library/Formula/sdl2.rb
+++ b/Library/Formula/sdl2.rb
@@ -1,8 +1,8 @@
 class Sdl2 < Formula
   desc "Low-level access to audio, keyboard, mouse, joystick, and graphics"
   homepage "https://www.libsdl.org/"
-  url "https://libsdl.org/release/SDL2-2.0.3.tar.gz"
-  sha256 "a5a69a6abf80bcce713fa873607735fe712f44276a7f048d60a61bb2f6b3c90c"
+  url "https://libsdl.org/release/SDL2-2.0.8.tar.gz"
+  sha256 "edc77c57308661d576e843344d8638e025a7818bff73f8fbfab09c3c5fd092ec"
 
   bottle do
     cellar :any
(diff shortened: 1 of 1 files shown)


miniupnp:
my OS X 10.5 patches : https://gist.github.com/miniupnp/6de808a8939711fcb856eaa049c18883

miniupnp:
hum doesn't work well :(

Wohlstand:
I made SDL 2.26 work on Mac OS X Tiger, but as an X11 application (because I was lazy to port the whole Cocoa counterpart): https://github.com/Wohlstand/SDL/tree/mac-tiger, but, I did Tiger-specific modules being separated of the main "apple" modules to maintain them easier and don't conflict to the central Apple part.

Library/Formula/sdl2.rb
```diff
@@ -26,8 +26,8 @@ class Sdl2 < Formula
   # https://github.com/mistydemeo/tigerbrew/issues/361
   if MacOS.version <= :snow_leopard
     patch do
-      url "https://gist.githubusercontent.com/miniupnp/26d6e967570e5729a757/raw/1a86f3cdfadbd9b74172716abd26114d9cb115d5/SDL2-2.0.3_OSX_104.patch"
-      sha256 "4d01f05f02568e565978308e42e98b4da2b62b1451f71c29d24e11202498837e"
+      url "https://gist.githubusercontent.com/miniupnp/6de808a8939711fcb856eaa049c18883/raw/1e4e0c111fb56c20905a9cd238323a298e1acbe0/SDL2-2.0.8_OSX_105.patch"
```
AJenbo, line 29: This fails on 10.4, i guess because the patch isn't targeting the older version of OSX :(

AJenbo:
Hm I would like to try and fix this, but I don't understand how on runs the build manually or have it leave the source so I can correct things between each run.

If I try and download the src and apply the patch i get the following errors when running `./autogen.sh --without-x --disable-haptic --disable-joystick`:
```
configure:275: error: possibly undefined macro: PKG_CONFIG_LIBDIR
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
Couldn't find autoconf, aborting
Now you are ready to run ./configure
```

And if i push ahead and run `./configure` it just ends with:
```
checking for Altivec with GCC altivec.h and -maltivec option... yes
./configure: line 26842: syntax error near unexpected token `else'
./configure: line 26842: `else'
```

And I'm unable to run `make` at that point.

sevan:
@AJenbo If you invoke `brew install` with `-d` you'll be given an option to drop to a shell in the build env. That's one way to work. Other is to apply the patch outside of brew to source which you have extracted, `autogen` and make use of the patches in the formula (whether remotely via a gist or in the file itself).

Do you know which functionality you require from SDL2 that's missing from v2.0.3?

AJenbo:
uploading every failed attempts at fixing a bug to github doesn't seem very efficient, i figured out the -d method though, that is workable so thanks for the hint.

AJenbo:
> hum doesn't work well :(

@minipnp could you elaborate on that?

AJenbo:
> hum doesn't work well :(

@miniupnp could you elaborate on that?

PR #597, opened 2018-07-07
Fix vendored Ruby script on G3 (again)

### LINKED ISSUE #593: [Hack] Fix for SHA256 mismatch on G3 hardware
First of all @mistydemeo et al, huge thanks for maintaining this fork! This project makes playing around with old PowerPC Macs a blast!

My setup:
G3 PowerBook3,1 (Firewire) 400MHz
Mac OS X 10.4.11

I've been having this weird issue where on every package/dependency/resource, the install fails due to a SHA256 mismatch.
I wasn't able to find the root cause for this, so I was forced to edit every formula and replace the hashes by hand.

Just today I hacked together a (probably terrible) and temporary solution to this problem: Disabling the SHA256 checks entirely.

I've never written a line of Ruby, so apologies for the crudeness of the implementation.

### To disable SHA256 checks, comment out the following lines with a `#` character at the beginning of the line.

`/usr/local/Library/Homebrew/resource.rb` line **80**
`/usr/local/Library/Homebrew/patch.rb` line **165**
`/usr/local/Library/Homebrew/formula_installer.rb` line **712**

I just put this together in case someone else out there is having the same problem. Please close this issue if there is a better solution I missed!

mistydemeo:
Fixes #593.

I was checking the CPU model incorrectly for PPC, which broke my attempt at guarding G3 and G4 from using Ruby 1.8.2's sha2. This fixes that check, and also fixes another bug that was hidden by this other bug. It's bugs all the way down!

cc @valtteri

### DIFF
--- a/Library/Homebrew/cmd/vendor-install.sh
+++ b/Library/Homebrew/cmd/vendor-install.sh
@@ -97,7 +97,12 @@ fetch() {
     trap - SIGINT
   fi
 
-  cpu_family="$(sysctl -n hw.cpufamily)"
+  if [[ "$(sysctl -n hw.cputype)" = "18" ]]; then
+    cpu_family="$(sysctl -n hw.cpusubtype)"
+  else
+    cpu_family="$(sysctl -n hw.cpufamily)"
+  fi
+
   if [[ -x "$(which shasum)" ]]
   then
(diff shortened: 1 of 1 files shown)


VKoskiv:
Great work! It appears to be working perfectly now, and it's great to know other G3 systems benefit from this fix.

mistydemeo:
Great! Glad it's working for you.

PR #598, opened 2018-07-07
Update Python and Python 3 formulae

mistydemeo:


w-och:
Re issue #599, this seems to cause pip3 to disappear

PR #602, opened 2018-07-20
Remove Tiger-only requirement for apple-gcc4.2

nglevin:
GCC 4.2 isn't offered through Xcode 3.1.4, the latest version of Xcode available for Leopard. That still shipped with Apple's GCC 4.0.1.

Xcode 3.2.x was the first to include GCC 4.2.x. Xcode 3.2 cannot be installed on any Mac OS X before Snow Leopard.

Unfortunately, this version check isn't very useful for Leopard users. Removing it, it installs and runs fine on Leopard PPC on a G4.

Happy to back away from this change, if there is indeed another means of having GCC 4.2 on Leopard minus the Snow. I couldn't find it on http://developer.apple.com/download/more/ .

### DIFF
--- a/Library/Formula/apple-gcc42.rb
+++ b/Library/Formula/apple-gcc42.rb
@@ -1,23 +1,11 @@
 require 'formula'
 
-class TigerOnly < Requirement
-  def message; <<-EOS.undent
-    gcc-4.2 is only provided for Tiger, as it is officially available
-    from an Xcode shipped with Leopard.
-    EOS
-  end
-  def satisfied?; MacOS.version == :tiger; end
-  def fatal?; true; end
-end
-
 class AppleGcc42 < Formula
   homepage 'http://r.research.att.com/tools/'
   url 'https://ia902307.us.archive.org/31/items/tigerbrew/gcc-42-5553-darwin8-all.tar.gz'
   mirror 'http://r.research.att.com/gcc-42-5553-darwin8-all.tar.gz'
   version '4.2.1-5553'
   sha1 '0e529a2e4723e016e3d086d6ca3215d700931503'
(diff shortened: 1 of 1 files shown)


mistydemeo:
Xcode 3.1.4 should include GCC 4.2; Xcode 3 was the first release to ship 4.2. That's the reason the package is marked Tiger-only right now.

I'm not sure what `/usr/bin/gcc` is, but there should be a `/usr/bin/gcc-4.2`. Do you not see that file?

nglevin:
Fascinating! So `/usr/bin/gcc` is gcc 4.0.1 on Leopard, but `/usr/bin/gcc-4.2` is... yes, GCC 4.2.1.

*sound of a truck swiftly backing up*

mistydemeo:
¯\_(ツ)_/¯ It's a weird decision on Apple's part! But Tigerbrew is configured to prefer GCC 4.2, regardless of what `/usr/bin/gcc` is.

nglevin:
I mean, it makes sense, from the standpoint of not breaking source-level compatibility if Leopard apps are being built assuming `gcc` is GCC 4.0.1, not 4.2.x. I won't knock Apple for that. :)

More so since Apple shipped a ton of sample code on the Leopard DVD assuming GCC 4.0.1, which is a nice surprise for learning Mac OS X development in an iOS-focused time.

I just got Leopard running on a PowerBook G4 and I'm having fun looking at `dtrace`, `libauto`, `zfs` (which only has write support when built from source, funny enough), Xcode 3 and all these other pre-clattner platform firsts. There's something really neat about swapping out and rebuilding all these little pieces without retaliation in an unsupported OS.

Issue #265, opened 2014-11-11
Whois 5.2.1 build fails (10.4/G4)


dcwar:
Previous version (5.2.0) built with no issues. When building we get the error: "Can't locate autodie.pm in @INC ...."

System Perl is v5.8.6. Would you suggest upgrading?

Make
https://gist.github.com/dcwar/8b847092856cda252e09

Brew Config
https://gist.github.com/dcwar/1095519145bb3392233c

PR #608, opened 2018-09-06
Update coreutils. Correct whois regression

dcwar:
Update coreutils to 8.30. Correct regression in whois formula (see Tigerbrew #265)

### DIFF
--- a/Library/Formula/coreutils.rb
+++ b/Library/Formula/coreutils.rb
@@ -1,23 +1,19 @@
 class Coreutils < Formula
   desc "GNU File, Shell, and Text utilities"
   homepage "https://www.gnu.org/software/coreutils"
-  url "https://ftpmirror.gnu.org/coreutils/coreutils-8.25.tar.xz"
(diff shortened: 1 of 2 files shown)


dcwar:
Hmm. I'm not sure how `whois` came to be at that path on both of my systems, or why updates seem to go there... but there they are. I've only ever upgraded through Tigerbrew, so I can only guess that something went sideways with a previous Tigerbrew upgrade. 

If you have any thoughts - or references I might refer to - on how to fix that, I'd be most appreciative. (I'm not strong with git, and don't want to lose the entire Tigerbrew install by accident.)

Once I've gotten that straightened out, I can give libidn a shot. I'll start an issue for that as I'm getting a weird note from Tigerbrew when I attempt that.

dcwar:
As a point of clarification, are Tigerbrew specific formula supposed to go in `Library/Formula`, with reference copies from Homebrew in `Library/Taps/homebrew/homebrew-core`? Just want to be sure that nothing is wrong with my installs. 

Can work on libidn2 next.

PR #609, opened 2018-09-07
Whois 5.3.2

dcwar:
Whois 5.3.2 and libidn2. Note that libidn2 does not currently build successfully due to test failures in libunistring (test-localename and test-fgetc). Which is particularly annoying after a long build.


dcwar:
Will revise for newer version of whois.

PR #614, opened 2018-10-09
Fix dpkg links

NotAlexNoyle:
Currently, tigerbrew attempts to download a version of dpkg that doesn't exist. This fixes the links to reference a version that does.


Traace:
sha256 - dpkg_1.18.24.tar.xz

d853081d3e06bfd46a227056e591f094e42e78fa8a5793b0093bad30b710d7b4


But it still doesn't build. Perl error is back, but I got perl installed...

`checking for perl >= 5.14.2... configure: error: cannot find perl >= 5.14.2`

PR #622, opened 2018-12-23
Fix lua

npyl:
I just found out (Leopard 10.5.2) that lua doesn't get installed because the patch doesn't exist anymore in the [website](http://lua-users.org/files/wiki_insecure/power_patches/5.2/lua-5.2.3-sig_catch.patch).

This should fix it.  I don't know if lua-5.3 is also affected, probably is.


ticky (approved):
The patch checksum checks out :)

mistydemeo (approved):

npyl:
I would not expect this to pass after 2 years :D

PR #629, opened 2019-02-05
Remove wine

ahyattdev:
Wine doesn’t run on ppc so we might as well remove it.


EricFromCanada:
Installs on 10.5 Intel. (Had to bump libgphoto2 first, though.)

ahyattdev:
Does TigerBrew have a way of making packages unavailable on specific architectures?

tkbagz08:
Can I get it to work for me

EricFromCanada:
> Does TigerBrew have a way of making packages unavailable on specific architectures?

See apple-gcc42 for a custom `Requirement` that allows only Tiger; this could be modified to instead check that `Hardware.cpu_type == :intel`.

mistydemeo:
A `depends_on :arch => :intel` should be able to handle it. Sorry this slipped my mind! Would you be able to add that to the top of the dependencies?

JL2210:
Don't forget about Winelib! It provides a MinGW-compatible compiler for building Windows apps into native executables.

ahyattdev:
@JL2210

ahyattdev:
Looks like it shouldn't be removed

PR #630, opened 2019-02-17
Fix issue for tigerbrew install fail on PPC OSX 10.4 as sha mismatched for ruby and curl

staycatz:
Current tigerbrew breaks due to sha mismatched on PPC Tiger (10.4) , changing vendor-install to match SHA and make it validate and install those package properly.

Tested on PPC Tiger 10.4 with PPC Mac mini


staycatz:
@mistydemeo You're right, look like it's ruby 1.8.2 sha256 bug. I'm closing this PR and raise another one to fix detection of Ruby 1.8.2 on ppc7450 which is my PPC Mac mini.

PR #631, opened 2019-02-17
Fix ruby 1.8.2 detection to skip ppc7450 cpu

staycatz:
This is added to skip Ruby 1.8.2 on OSX 10.4 running on PPC7450 (Mac mini G4)
on Mac mini G4, the cpu_type is 18 but cpu_family is 11, and will cause failure as Ruby buggy sha256 logic will be used.

### DIFF
--- a/Library/Homebrew/cmd/vendor-install.sh
+++ b/Library/Homebrew/cmd/vendor-install.sh
@@ -109,7 +109,7 @@ fetch() {
   elif [[ -x "$(which sha256sum)" ]]
   then
     sha="$(sha256sum "$CACHED_LOCATION" | cut -d' ' -f1)"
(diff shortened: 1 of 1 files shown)


Library/Homebrew/cmd/vendor-install.sh
```diff
@@ -110,7 +110,7 @@ fetch() {
   then
     sha="$(sha256sum "$CACHED_LOCATION" | cut -d' ' -f1)"
   # Ruby 1.8.2's vendored Ruby has broken SHA256 calculation pre-G4e
```
mistydemeo: Thank you! Probably makes sense to update this comment, too:

```suggestion
  # Ruby 1.8.2's vendored Ruby has broken SHA256 calculation on several PowerPC CPUs
```

mistydemeo:
Thanks again!

PR #634, opened 2019-03-15
--update to libwebsockets

tree-s:
Up dating the libwebsockets Formula to the latest version compatible with Legacy OSX


mistydemeo (approved):
Haven't been able to test, but looks good.

PR #641, opened 2019-07-21
Update formula for Emacs 26.2

lubert:
- Tested on an iBook G4 running 10.4.11
- The `configure` patch shouldn't affect x86 builds, as it is more permissive

### DIFF
--- a/Library/Formula/emacs.rb
+++ b/Library/Formula/emacs.rb
@@ -1,9 +1,9 @@
 class Emacs < Formula
   desc "GNU Emacs text editor"
   homepage "https://www.gnu.org/software/emacs/"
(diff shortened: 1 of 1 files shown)


Library/Formula/emacs.rb
```diff
@@ -43,6 +43,8 @@ class Emacs < Formula
   depends_on "mailutils" => :optional
   depends_on "glib" => :optional
 
+  patch :DATA
```
mistydemeo: Do you mind adding a comment about what this does?

lubert: Good call, will do 👍 !

lubert:
@mistydemeo added a comment, any chance of getting this and https://github.com/mistydemeo/tigerbrew/pull/642 merged soon? Thank you!

PR #647, opened 2019-10-28
Using Perl SHA256 if if shasum/sha256sum unavailable

chapmajs:
First, thanks for a great project! It's incredible that folks are still supporting OS X 10.4!

I did a clean install of OS X 10.4 on an ALS iMac G5 today, updated to 10.4.11, installed XCode 2.5, and tried to bootstrap `tigerbrew`. I ran into issues with `brew doctor` after initial download: I couldn't install the vendored Ruby or cURL due to a SHA checksum mismatch. I tracked this down to the `install-vendor.sh` script, which had a conditional to first look for OS-provided SHA binaries, then try a Ruby implementation if that failed, or use a Perl implementation if we're on a specific subset of PPC chips.

Apparently the G5 in the ALS iMacs (970/970fx) are also one of the CPUs that Ruby 1.8.2 has a bad SHA256 on. My `sysctl -n hw.cpusubtype` is 100, which is not in the excluded list. Rather than trying to figure out every subtype that has a broken SHA256, why don't we just always use the Perl implementation?


chapmajs:
OK, makes sense to me! I updated the CPU exclusion list to exclude the 970/970fx and force-pushed the change to my `sha256_fix` branch.

PR #649, opened 2019-11-09
Update url in fortune.rb

jupo42:
The source download requires having an FTP URL instead of an HTTP URL to fetch the file.


ablakely:
Can confirm same issue.

PR #650, opened 2019-11-10
Update url for icu4c.rb

jupo42:
Source download is now on SourceForge and mirror listed is no longer valid. Updated to current URL.


jupo42:
Now superceded by PR #737 which includes the mirror as noted above and corrected HEAD url.

PR #652, opened 2019-11-19
libtorrent: fix building with gcc-7.3.0

ralovich:
Instrumentation generates the following missing symbols on g4:
undefined reference to __sync_add_and_fetch_8

See also https://github.com/rakshasa/rtorrent/issues/156.


kencu:
gcc should automatically add -latomic to the link libraries, but it doesn't. This is probably a bug, and most upstream consider it to be so. If you want the instrumentation, you can manually add the library to your link line, and then the symbols will be found.

PR #662, opened 2020-02-29
graphviz.rb: update url (sha256 does not change)

berryamin:


cellularmitosis:
This was fixed in https://github.com/mistydemeo/tigerbrew/pull/680 and can be closed.

PR #667, opened 2020-03-23
Updated URL for popt.rb

pk2061:
I updated the URL in the formula for popt as old the url didn't seem to be working anymore.

Best regards
pk2061


jupo42:
It looks like the current project is being actively developed in the rpm repository on github. I would recommend changing homepage to "https://github.com/rpm-software-management/popt" and the download url to "https://github.com/rpm-software-management/popt/archive/popt-1_16-release.tar.gz" as the most official links available.

PR #683, opened 2020-05-13
Initial support for Panther

larb0b:
This adds initial support for Panther.

One would be able to `curl -fsSkL raw.github.com/mistydemeo/tigerbrew/go/bootstrap_panther | sudo sh - && ruby -e "$(curl -fsSkL raw.github.com/mistydemeo/tigerbrew/go/install)"` and tigerbrew should be installed. However, there are still issues as all of the ruby scripts in tigerbrew are shebanged with `/usr/bin/ruby `and not `/usr/bin/env ruby`. Working past this, there are still issues (fails to install vendor `ruby`), but this is a start.


larb0b:
Note that even with this `brew` still seems to want to use `/usr/bin/ruby` at least when it's initially run.

larb0b:
I'm now able to get as far as

```
/usr/local/bin/brew: line 219:  7906 Trace/BPT trap          "./$VENDOR_VERSION/bin/$VENDOR_NAME" --version >/dev/null 2>&1
Error: Failed to install vendor ruby.
Error: Failed to install vendor Ruby.
```

if I symlink `/usr/local/bin/ruby` to `/usr/bin/ruby`.

mistydemeo:
This is amazing, thank you! 🙌🏻

> I'm now able to get as far as

The vendor Ruby is built for Tiger, so I'm not surprised it doesn't run on Panther. We might have to build a new Ruby just for Panther in order to get something that runs.

There might be a few shebangs that need work to ensure everything runs using the vendored Ruby. This wasn't a huge issue on Tigerbrew because a lot of the most basic code has been written to support 1.8.2, but I think that stuff should get fixed up for sure.

larb0b:
> There might be a few shebangs that need work to ensure everything runs using the vendored Ruby.

Yeah, I only took a cursory look at everything (I'm still not entirely sure how `brew` picks which `ruby` it's using before vendor Ruby installed), but it seems most things (at least in ENV) use `/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby`.

Maybe we could just have this install script to just install a Ruby as vendor Ruby for `brew` as a lazy hack?

> The vendor Ruby is built for Tiger, so I'm not surprised it doesn't run on Panther. We might have to build a new Ruby just for Panther in order to get something that runs.

What version of Ruby would you want minimum for the vendor Ruby? There were some errors when I tried to build 2.7.1 (and it also requires that you build a newer GNU Make), but I'm sure they could be patched around if needed.

larb0b:
I tried doing it with 1.8.5 (https://github.com/larb0b/tigerbrew/blob/master/Library/Homebrew/cmd/vendor-install.sh) but 

```
$ brew
Error: Failed to install vendor ruby.
Error: Failed to install vendor Ruby.
```

it'd be nice to get more verbosity out of this!

larb0b:
https://github.com/larb0b/tigerbrew/blob/master/Library/Homebrew/utils/ruby.sh

This workaround works to get it around trying to install a vendor Ruby but 1.8.5 is too old to work for brew. There are numerous errors when compiling later versions on Panther, however.

mistydemeo:
> What version of Ruby would you want minimum for the vendor Ruby? There were some errors when I tried to build 2.7.1 (and it also requires that you build a newer GNU Make), but I'm sure they could be patched around if needed.

We're currently using Ruby 2.3, though I'll be upgrading later. What were the errors?

> By the way, I can supply SSH access to a Panther machine if any contributors would like to help.

That'd be amazing! I currently don't have any hardware able to run Panther that could work on this.

larb0b:
I don't remember what the errors were at this point and I don't think I have them saved anywhere, but I might try to reproduce my progress at some point.

I've set up a 1.33 GHz iBook G4 w/ the latest OpenSSH, if you send a key my way then I can provide access.

PR #699, opened 2020-07-29
Update dcraw.rb

Ricotan-naboon:
Fix Doesn't Install Issue.
Checking OS X 10.7.5(11G63) on iMac 5,1 and Mac Pro 2,1.


Ricotan-naboon:
Because URL is Dead.

PR #701, opened 2020-09-03
Add Xcode links and information to Readme

ticky:
This adds the required Xcode versions for Tiger and Leopard to the Readme.

These links are more current than the https://developer.apple.com/xcode/downloads/ url currently emitted `brew doctor` (which redirects to Apple’s beta software downloads), and each list a single Xcode version in their results pane. At least until Apple change the format again! 😅

image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ A little experimental fork of [Homebrew][homebrew] that adds support for PowerPC
 Installation
 ============
 
+You will first need the newest version of Xcode for your operating system installed. For Tiger that's [Xcode 2.5, available from Apple here](https://developer.apple.com/download/more/?=xcode%202.5). For Leopard, [Xcode 3.1.4, available from Apple here](https://developer.apple.com/download/more/?=xcode%203.1.4). Both downloads will require an Apple Developer account.
+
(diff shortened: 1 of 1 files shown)


ticky:
I’d also love to follow up and add more specific URLs to the `brew doctor` output, if that would be desired!

https://github.com/mistydemeo/tigerbrew/blob/da9c583c2de859d229c130c7a81482ed0c9aa702/Library/Homebrew/cmd/doctor.rb#L323-L341

And possibly also [`BuildToolsError`](https://github.com/mistydemeo/tigerbrew/blob/da9c583c2de859d229c130c7a81482ed0c9aa702/Library/Homebrew/exceptions.rb#L263) and [`BuildFlagsError`](https://github.com/mistydemeo/tigerbrew/blob/da9c583c2de859d229c130c7a81482ed0c9aa702/Library/Homebrew/exceptions.rb#L310).

mistydemeo (approved):
Thank you!

mistydemeo:
> I’d also love to follow up and add more specific URLs to the brew doctor output, if that would be desired!

That'd be great!

PR #702, opened 2020-09-03
Point links in contribution guide at Tigerbrew

ticky:
These links went to the upstream Homebrew project, and some of them are no longer at that URL. This updates them to point at this project’s copy.


mistydemeo (approved):
Thank you!

PR #703, opened 2020-09-03
Un-quieten vendor installs

ticky:
The vendor installation and subsequent checksum verification of curl and ruby takes quite a long time on older machines; somewhere in the order of an hour on my PowerBook G4 Titanium’s 867 MHz CPU and ailing disk drive.

For that entire time, unless you’ve set Tigerbrew to verbose, you’re given _absolutely no_ output until the “real” Ruby infrastructure takes over, which is easy to mistake for nothing actually happening!

For that reason, and having talked to @mistydemeo about it in person, I’d propose removing the `--quiet` flags from these steps, so that `vendor-install` _at least_ goes and prints that it’s downloading stuff.

This way there’s some feedback almost immediately that it’s understandable could take some time, which IMO makes for a better first-launch experience on older machines. 👍🏼

### DIFF
--- a/Library/Homebrew/cmd/vendor-curl.sh
+++ b/Library/Homebrew/cmd/vendor-curl.sh
@@ -15,7 +15,7 @@ setup-curl-path() {
 
       if [[ $(readlink "$vendor_curl_current_version") != "$(<"$vendor_dir/portable-curl-version")" ]]
       then
-        if ! brew vendor-install curl --quiet
+        if ! brew vendor-install curl
         then
           onoe "Failed to upgrade vendor Curl."
         fi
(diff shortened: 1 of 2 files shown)


mistydemeo:
How do you feel about this compared to the message that

mistydemeo:
How do you feel about this compared to the message that's printed as of #718? Would you like to see both?

ticky:
I think it could go either way, but I liked that this reveals more readily that stuff was progressing (i.e. it notes it was downloading curl, then Ruby, so at least it wasn't so easily confused for being hung). But at least printing _something_ early makes it clear that it hasn't just gotten stuck!

mistydemeo (approved):

PR #724, opened 2021-02-27
Added mirror for PCRE because the two existing links did not work for me 

pk2061:
Hi,

I created a new mirror for the PCRE to https://ftp.pcre.org/pub/pcre/pcre-8.39.tar.bz2 because the existing links seemed to be broken.

### DIFF
--- a/Library/Formula/popt.rb
+++ b/Library/Formula/popt.rb
@@ -1,7 +1,7 @@
 class Popt < Formula
   desc "Library like getopt(3) with a number of enhancements"
   homepage "http://rpm5.org"
-  url "http://rpm5.org/files/popt/popt-1.16.tar.gz"
+  url "https://fossies.org/linux/misc/popt-1.16.tar.gz"
   sha256 "e728ed296fe9f069a0e005003c3d6b2dde3d9cad453422a10d6558616d304cc8"
 
   bottle do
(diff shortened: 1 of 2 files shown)


Library/Formula/popt.rb
```diff
@@ -1,7 +1,7 @@
 class Popt < Formula
   desc "Library like getopt(3) with a number of enhancements"
   homepage "http://rpm5.org"
-  url "http://rpm5.org/files/popt/popt-1.16.tar.gz"
+  url "https://fossies.org/linux/misc/popt-1.16.tar.gz"
```
mistydemeo, line 4: Looks like this is a 404.

mistydemeo (changes requested):

Library/Formula/popt.rb
pk2061: Sorry, I did not mean to upload that as is.
I will take care of that.. :-)

pk2061: Maybe just discard the changes to the popt.rb as here the original url http://rpm5.org/files/popt/popt-1.16.tar.gz seems to work just fine.
Sorry for the mess-up on my end here!

mistydemeo: No worries at all!

mistydemeo:
Merged; thank you very much for your contribution!

cozycactus:
pcre links still broken here...

mistydemeo:
This mirror worked when this PR was merged, but has broken since. I'll find a new source.

PR #731, opened 2021-05-03
jpeg: correct SHA256 checksum

LegoLivesMatter:
For an unknown reason, the SHA256 checksum of [https://ijg.org/files/jpegsrc.v8d.tar.gz](jpegsrc.v8d.tar.gz) has changed, and as a result jpeg fails to install due to a checksum mismatch. This PR corrects that issue.


evanmiller:
I don't see this new checksum documented anywhere on the Internet, but I can confirm that I compute the same one on multiple machines using different programs (so it's likely not a duplicate of #647).

It's strange that the JPEG authors would change the file.

mistydemeo (approved):
Thanks! Sorry I missed this until now.

PR #733, opened 2021-05-30
Expat: Update URLs and HEAD

evanmiller:
Expat 2.1.0 is deprecated and has been removed from a number of mirrors. Restore working links until support for a more recent version of expat is added.


mistydemeo (approved):
Thanks!

PR #735, opened 2021-06-09
Use newer make for projects with GObject introspection

evanmiller:
On Tiger, building `atk` and `gdk-pixbuf` with the system-provided `make` results in errors such as:

```
  *** Need to define Atk_1_0_gir_LIBS or Atk_1_0_gir_PROGRAM.  Stop.
```

Using the brew-provided `make` fixes the issue.

### DIFF
--- a/Library/Formula/atk.rb
+++ b/Library/Formula/atk.rb
@@ -11,6 +11,7 @@ class Atk < Formula
   end
 
   depends_on "pkg-config" => :build
+  depends_on "make" => :build if MacOS.version < :leopard
   depends_on "glib"
   depends_on "gobject-introspection"
 
@@ -21,8 +22,8 @@ class Atk < Formula
     system "./configure", "--disable-dependency-tracking",
                           "--prefix=#{prefix}",
                           "--enable-introspection=yes"
-    system "make"
-    system "make", "install"
+    system make_path
+    system make_path, "install"
   end
 
   test do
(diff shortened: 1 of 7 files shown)


evanmiller:
See also #481 - probably these formulas should use a standard solution... I believe it's sufficient to change

    depends_on "gobject-introspection"

to

    depends_on "gobject-introspection" => :build

And then `system "make"` to `make`. The `gobject-introspection` formula has brew's `make` as a build-time requirement, so the formulas won't need to require `make` themselves.

ticky (approved):
I've just tested this while attempting to confirm the fixes in #738, and along with the changes in #745, this works.

ticky:
Oh, one comment is that `gtk+` is [still marked as requiring Leopard](https://github.com/mistydemeo/tigerbrew/blob/d28bbe7cd204048e48639e2e37b407bb588cf5f0/Library/Formula/gtk%2B.rb#L23), I haven't yet built that on Tiger, but I presume this means you tested that?

ticky:
Oh, and the dependency on `make` via `gobject-introspection` doesn't appear to work on my machine unless explicitly declared in these dependencies. From what I can tell we need both, unfortunately.

ticky:
`gtk+` seems to need similar changes to #738 around `NSTextInputClient`. I've yet to find any `gtk+3` software which doesn't transitively depend on `gtk+` - is there any? In any case this isn't a showstopper, just means there's more to do :)

mistydemeo (approved):

PR #738, opened 2021-06-09
gtk+3 fixes on 10.4

evanmiller:
Here are a few changes that I found necessary to compile the gtk+3 formula on 10.4.11. They are workarounds for the following facts:

* Tiger has CUPS 1.1 but not 1.2
* Tiger doesn't have the `NSTextInputClient` protocol. I am guessing this was an informal protocol prior to 10.5; it's possible that removing the protocol declaration will have a run-time effect, but in my experience Cocoa frameworks check for individual methods rather than protocol conformance. Anyway, compiling is better than not.
* Tiger's default linker doesn't find Carbon symbols (e.g. `GetCurrentButtonState`) without additional flags

### DIFF
--- a/Library/Formula/gtk+3.rb
+++ b/Library/Formula/gtk+3.rb
@@ -24,6 +24,8 @@ class Gtkx3 < Formula
   depends_on "glib"
   depends_on "hicolor-icon-theme"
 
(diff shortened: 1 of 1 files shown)


ticky (approved):
With #735 and #745 applied, this works.

mistydemeo (approved):

PR #740, opened 2021-06-21
sleepwatcher: Add older version which works on Tiger

ticky:
SleepWatcher 2.1 and newer don't work on Tiger, but 2.0 is still provided. It requires the use of a startup item rather than launchd like the newer versions, so it's a bit different.


mistydemeo (approved):
Thank you!

PR #743, opened 2021-06-26
jpeg: Update to v9d

ticky:
jpeg has been updated several times since v8d in 2012, with the latest in early 2020. This updates us to the latest, which builds and runs just fine on Leopard. I've still got to test Tiger, but it seems fairly likely to be portable.


mistydemeo (approved):
Doesn't look like this required mass rebottling in Homebrew.

ticky:
Just confirmed this works fine on Tiger, too! :)

PR #747, opened 2021-07-10
graphicsmagick: Update download URL, as it has moved

ticky:


mistydemeo (commented):
Looks like this is a different sha256. Mind updating that too?

ticky:
Sorry, I couldn't check while a download was running, given you can't curl a sourcegorge URL 😫

It's updated with the valid URL and SHA now, confirmed it builds on both Tiger and Leopard 👍🏼

mistydemeo (approved):

PR #759, opened 2021-10-12
curl-ca-bundle 2021-09-30

mistydemeo:


mistydemeo:
Confirmed this works. Next step is to update the vendored curl.

PR #767, opened 2021-10-15
ENV (superenv / stdenv): Mark every future version of GCC above gcc-4.8 (which includes the as-of-current-writing gcc-7) as supporting C++11, which fixes `ENV.cxx11` usage in Tigerbrew formulae.

akemin-dayo:
The previous version of the regular expression only detected `gcc-4.8`, `gcc-4.9`, and `gcc-5` as supporting C++11, but erroneously fails to detect future versions as supporting C++11 (presumably because upstream assumes you'll be using.

This currently results in any Tigerbrew formulae using `ENV.cxx11` to incorrectly throw an error when a user has the latest version of the `gcc` formula installed (which installs `gcc-7` as of this writing).

The updated regular expression successfully detects not only `gcc-7`, but also future versions all the way up to `gcc-99`(!) for futureproofing purposes, in case the `gcc` formula gets updated to a newer version of GCC at some point in the future.


sevan:
Just tried to use `ENV.cxx11` in a formula with GCC 7 installed and needed the patches here before it would work.

PR #781, opened 2022-01-04
dirmngr: fix patch URL

### LINKED ISSUE #780: Fix patch URL in dirmngr formula
Looks like the version of Homebrew that tigerbrew was forked from had this exact same bug because the gnupg maintainers changed their hosting of the patch in question so it's no longer at that URL.

https://github.com/Homebrew/homebrew-core/issues/12001
https://github.com/Homebrew/homebrew-core/pull/12005

I ran into this when I was trying to brew gpg2, which depends on dirmngr. I have found the fix to the original homebrew-core issue, and I can quickly bring it over to the tigerbrew fork.

bleungatchromium:
Closes #780.

Signed-off-by: Alyssa Ross <hi@alyssa.is>
Signed-off-by: Benson Leung <bleung@chromium.org>


mistydemeo (approved):
Thank you!

Issue #779, opened 2021-12-07
socat failed to download


996refuse:
it looks this link won't work anymore.
http://www.dest-unreach.org/socat/download/socat-1.7.3.0.tar.gz

Ricotan-naboon:
Did you did edit socat.rb?
I think fix URL it will work.

http://www.dest-unreach.org/socat/download/Archive/socat-1.7.3.0.tar.gz

sevan:
Updated url was committed to Tigerbrew.

PR #788, opened 2022-07-05
Update socat.rb

Ricotan-Linus:
#779 
I didn't check it will work(My Old PC was broken), plz check before marge...


Ricotan-Linus:
Ok, I check Fix that.

mistydemeo (approved):

mistydemeo:
Thank you!

PR #789, opened 2022-07-11
Fixed neofetch.rb and added xscreensaver.rb

ablakely:
Tested with a fresh install of Tiger and it failed trying to copy to /usr/bin, fixed by setting prefix.

Not sure why this wasn't happening when testing before.


mistydemeo:
Thanks! Applied the patch.

You might find it easier opening pull requests from alternate branches instead of `master`; that makes keeping up with Tigerbrew's `master` a bit easier. 😄

PR #793, opened 2022-09-04
the_silver_searcher: Update to 0.33.0 and fix build on Tiger

jorio:
The current formula attempts to build version 0.31.0, but fails on Tiger.
This PR updates the_silver_searcher to 0.33.0 and fixes the build on Tiger.
Updating past 0.33.0 would require more substantial patches.


mistydemeo (approved):
Thank you!

PR #805, opened 2023-05-16
openssl: Upgrade formula to v1.1.1u

sevan:
Build tested on Tiger PowerPC
With this change `brew audit openssl` has one error flagged which is `Header files that shadow system header files were installed to "tigerbrew/Cellar/openssl/1.1.1t/include"` but unsure how to address it. Pointers appreciated if this needs to be addressed before it can be merged.

### DIFF
--- a/Library/Formula/openssl.rb
+++ b/Library/Formula/openssl.rb
@@ -1,15 +1,9 @@
 class Openssl < Formula
   desc "SSL/TLS cryptography library"
   homepage "https://openssl.org/"
-  url "https://www.openssl.org/source/openssl-1.1.1t.tar.gz"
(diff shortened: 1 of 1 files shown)


sevan:
Looks like it does need to be addressed.

sevan:
Now just need need to clean up the fallout from the API change since OpenSSL 1.1.0 for anything that's not up to date. :)

goshhhy:
i'm seeing the same behavior - TigerBrew does not work out of the box due to test failures in OpenSSL which prevent the installation of curl and git. manually installing this recipe works around the issue successfully.

mistydemeo:
Building now. Given the number of other packages that will need to be rebuilt, I'll try to coordinate merges.

mistydemeo:
Merged into the OpenSSL 1.1 branch.

PR #807, opened 2023-05-17
libevent: Upgrade formula to v2.1.12-stable

sevan:
Needed to support OpenSSL version 1.1.0 and newer.


mistydemeo (approved):
Looks like this'll break tmux - I'll get that revision bumped and bottled in the openssl staging branch.

PR #809, opened 2023-05-17
zlib: Upgrade formula to v1.2.13

sevan:

PR #811, opened 2023-05-17
curl: Upgrade formula to v8.1.2

sevan:
libmetalink is no longer supported

Uses updated zlib from tigerbrew so depends on PR #809


goshhhy:
this works and passes audit for me on Leopard on a G4, alongside the mentioned PRs.

it is also required, along with associated PRs, for me to use tigerbrew successfully on that system.

mistydemeo (approved):

PR #812, opened 2023-05-17
mercurial: Upgrade formula to v6.5.1

sevan:


mistydemeo (approved):
Thank you! I'm a little impressed it still builds.

sevan:
Bizarrely had the exact same big endian issue as git via sha1dc which both use.

PR #814, opened 2023-05-17
apr: Upgrade formula to v1.7.4

sevan:
Seems to build ok now without deparallelize.

### DIFF
--- a/Library/Formula/apr.rb
+++ b/Library/Formula/apr.rb
@@ -4,18 +4,13 @@ class Apr < Formula
   url "https://www.apache.org/dyn/closer.cgi?path=apr/apr-1.7.4.tar.bz2"
   sha256 "fc648de983f3a2a6c9e78dea1f180639bd2fad6c06d556d4367a701fe5c35577"
 
-  bottle do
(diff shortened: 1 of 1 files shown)


Library/Formula/apr.rb
```diff
@@ -1,24 +1,15 @@
...
   keg_only :provided_by_osx, "Apple's CLT package contains apr."
 
   option :universal
 
   def install
     ENV.universal_binary if build.universal?
-
-    # https://bz.apache.org/bugzilla/show_bug.cgi?id=57359
-    # The internal libtool throws an enormous strop if we don't do...
     ENV.deparallelize
```
mistydemeo, line 13: Did you mean to delete this in addition to the comment?

mistydemeo (approved):

Library/Formula/apr.rb
sevan: Yes, I didn't think it through, url is no longer valid, comment is still relevant as I did hit issues with libtool.
I received error from libtool that "not a valid libtool object" but the build log in the bug report doesn't show that.
Is it just because the log didn't capture stderr and it is the same issue?

PR #818, opened 2023-05-17
subversion: Upgrade formula to v1.14.2

sevan:
Along with serf 1.3.9
There currently isn't a formula for utf8proc so the internal implementation is used.

Ideally should be merged after PR #814 #815 to upgrade apr and apr-utils, #816 #817 are requirements.
With these updates, it's possible to build subversion on Tiger with stock GCC 4.0, avoiding linker/compiler upgrade.


mistydemeo (approved):
Thank you!

PR #825, opened 2023-05-20
xxhash: Add formula for v0.8.1

sevan:
Adapted from
https://github.com/Homebrew/homebrew-core/blob/a78cbe4ff3223795b80705287c53eebe602b7c85/Formula/xxhash.rb


mistydemeo (approved):
Thank you!

PR #826, opened 2023-05-20
rsync: Add formula for v3.2.7

### LINKED ISSUE #791: rsync failed to build on 10.4
https://gist.github.com/happy-dreame/aaf49c4387dd4e0bb188b058aa97bcf6

Thank you.

sevan:
Adapted from
https://github.com/Homebrew/homebrew-core/blob/a76c6520515986d4c77bab61cab99076bcc2b727/Formula/rsync.rb

Resolves #791 
Need PR #825 to build, and ideally #824

### DIFF
--- a/Library/Formula/rsync.rb
+++ b/Library/Formula/rsync.rb
@@ -1,0 +1,1385 @@
+class Rsync < Formula
+  desc "Utility that provides fast incremental file transfer"
+  homepage "https://rsync.samba.org/"
+  url "https://rsync.samba.org/ftp/rsync/rsync-3.2.7.tar.gz"
+  mirror "https://mirrors.kernel.org/gentoo/distfiles/rsync-3.2.7.tar.gz"
+  mirror "https://www.mirrorservice.org/sites/rsync.samba.org/rsync-3.2.7.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/rsync.rb
```diff
@@ -0,0 +1,1385 @@
...
+  depends_on "lz4"
+  depends_on "openssl"
+  depends_on "popt"
+  depends_on "xxhash"
+  depends_on "zlib"
+
+  def install
+    ENV.append_to_cflags "-DSUPPORT_FILEFLAGS"
+    args = %W[
+      --prefix=#{HOMEBREW_PREFIX}
```
mistydemeo, line 19: Can you make this `prefix` instead? This is installing rsync in a way where it's not being managed by Tigerbrew. Unfortunately we don't have the sandboxing setup to actively prevent that right now.

mistydemeo (changes requested):

Library/Formula/rsync.rb
sevan: Sure, don.

mistydemeo (approved):
Thank you!

PR #829, opened 2023-05-21
readline: Upgrade formula to v8.2

sevan:


Library/Formula/readline.rb
```diff
@@ -21,6 +13,9 @@ class Readline < Formula
 
   def install
     ENV.universal_binary
+    # Since we don't set any CFLAGS on Intel, readline adds some
+    # which break the build as they're not supported by GCC 4.2
+    ENV.append_to_cflags "-g -O2" if Hardware::CPU.type == :intel
```
mistydemeo, line 18: Huh, interesting failure case. Thanks for fixing this!

mistydemeo (approved):

PR #830, opened 2023-05-21
perl: Upgrade formula to v5.38.0

sevan:


mistydemeo:
I'm testing this now, and will build Tiger G4/G5 bottles. It may take me a bit since some of the PRs are big packages.

mistydemeo:
`make test` failed for me when building for G4 on G5; testing now with a native G5 build.

mistydemeo:
Huh. I can't even get the *old* perl to build right. Out of curiosity, what OS and CPU are you testing on?

mistydemeo:
Oh, I see the problem. Today, it runs `make test` only when bottling (or when `--with-test` is passed). You probably didn't use that, so you didn't see it.

If it runs, and we know that the tests were already failing, I think this is fine. I can disable `make test` for now, and build binary packages for a few architectures.

sevan:
> Huh. I can't even get the _old_ perl to build right. Out of curiosity, what OS and CPU are you testing on?

Everything was built on a G5 iMac running on Tiger 10.4.11.

sevan:
Forgot to include one patch for the test suite, with this it should pass the testsuite, right?
https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/pull/446/files

If you're hitting an issue with  cpan/Test-Simple/t/Test2/behavior/ipc_wait_timeout.t, that's a bug that needs to be fixed upstream ([problem is there with blead](https://perl5.test-smoke.org/report/5034141))

sevan:
@mistydemeo  There were some locale related changes that landed after Perl 5.36.1, in time for v5.38.0, which restricted the assumption about locale support to newer operating system (related to POSIX 2008 I think).
With Perl v5.38.0 the test suite runs without failure or noise on ppc OS X Tiger 10.4.11 when I invoke `brew install -v --with-tests perl`

mistydemeo (approved):

mistydemeo:
Thank you! Built G4e and G5 bottles.

Issue #746, opened 2021-06-28
librsvg: can't be built with system libxml2 on Tiger


ticky:
Tiger's built-in `libxml2` 2.6 is too old for `librsvg`, which wants a version newer than 2.7

ticky:
`librsvg` also doesn't seem to work even with that worked around, due to some error it won't elaborate on:

```
/usr/bin/ld: Undefined symbols:
_gzdirect
```

sevan:
Should be resolved with the new libxml2 version.

sevan:
This issue is resolved if you use an up to date copy of the Tigerbrew tree. Thanks for the bug report.

PR #833, opened 2023-05-22
libxml2: Upgrade formula to v2.11.4

sevan:
With zlib formula built as a shared library there is no linking issue
Resolves undefined symbol issue part reported in #746 

Requires PR #809 so that a shared library is available to link against.

### DIFF
--- a/Library/Formula/libxml2.rb
+++ b/Library/Formula/libxml2.rb
@@ -1,18 +1,8 @@
 class Libxml2 < Formula
   desc "GNOME XML library"
   homepage "http://xmlsoft.org"
-  url "http://xmlsoft.org/sources/libxml2-2.9.2.tar.gz"
-  mirror "ftp://xmlsoft.org/libxml2/libxml2-2.9.2.tar.gz"
-  sha256 "5178c30b151d044aefb1b08bf54c3003a0ac55c59c866763997529d60770d5bc"
-
-  bottle do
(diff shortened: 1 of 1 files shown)


Library/Formula/libxml2.rb
```diff
@@ -45,7 +30,8 @@ def install
     system "./configure", "--disable-dependency-tracking",
                           "--prefix=#{prefix}",
                           "--without-python",
-                          "--without-lzma"
+                          "--with-lzma=#{Formula["xz"].opt_prefix}",
+                          "--with-zlib=#{Formula["zlib"].opt_prefix}"
```
mistydemeo, line 34: This'll need `zlib` added as a dependency.

Library/Formula/libxml2.rb
```diff
@@ -28,12 +18,7 @@ class Libxml2 < Formula
 
   option :universal
 
-  depends_on :ld64
-
-  fails_with :llvm do
-    build 2326
-    cause "Undefined symbols when linking"
-  end
+  depends_on "xz" => :build
```
mistydemeo, line 21: This isn't a build dep; it links dynamically against it, so it'll also be needed at runtime.

mistydemeo (commented):
Thanks! I'll make the changes I mentioned after merge, when I add the bottle.

PR #834, opened 2023-05-23
expat: Upgrade formula to v2.5.0

sevan:
--disable-debug is unrecognised.


mistydemeo (approved):
Thankfully doesn't seem to have broken compatibility.

PR #835, opened 2023-05-23
cmake: Upgrade formula to v3.9.6

sevan:
From v3.10 onwards cmake enforces the requirement of a C++11 compiler. Python modules have been upgraded to latest version to support Python 2.7 or most recent version required by cmake itself e.g docutils. Switched cmake build to use libraries provided via Tigerbrew. Given up to date formulas, the build of cmake itself is faster since it skips on building its own private copies of libraries which are now quite old as well.


mistydemeo (approved):
Thank you! I know this was a big one to get updated.

PR #836, opened 2023-05-23
pcre: Upgrade formula to v8.45

sevan:
PCRE is now EoL (pcre2 is where development happens), so drop support from building from head of branch.


mistydemeo (approved):
Thank you!

PR #838, opened 2023-05-23
freetype: Upgrade formula to v2.13.0

sevan:


mistydemeo (approved):
Looks like they've dropped `freetype-config`; I've pushed a commit to restore it.

PR #846, opened 2023-05-25
gobject-introspection: Upgrade formula to v1.60.2

sevan:
v1.60.2 is the last version to support autoconf.
More recent versions require meson.


mistydemeo:
Hm, looks like this won't build without a glib upgrade. Seems it needs at least 2.58.0.

sevan:
Ah, ok. I'm running a WIP of glib 2.59.0 here so that's why it passed. Will try and finish it off and raise a pull requests in the next few days.

sevan:
PR #989 brings in glib 2.59.0, that should unblock this and the other PRs for merging (#847, #860)

mistydemeo (approved):
Built! Thanks so much again for this.

PR #850, opened 2023-05-25
s-lang: Upgrade formula to v2.3.3

sevan:


Library/Formula/s-lang.rb
```diff
@@ -20,7 +12,8 @@ def install
     png = Formula["libpng"]
     system "./configure", "--prefix=#{prefix}",
                           "--with-pnglib=#{png.lib}",
-                          "--with-pnginc=#{png.include}"
+                          "--with-pnginc=#{png.include}",
+                          "--with-z=#{Formula["zlib"].opt_prefix}"
```
mistydemeo, line 16: Looks like this isn't specified as a dependency - mind adding that?

mistydemeo (approved):

PR #854, opened 2023-05-26
pixman: Upgrade formula to v0.42.2

sevan:
TODO:
The code in the test section works, and completes successfully. The test is considered a failure by brew though as pixman_image_unref() returns a boolean value where true is successful, and brew expects a zero (false).


mistydemeo (approved):
Thanks! I'll fix the test in a followup commit.

PR #855, opened 2023-05-30
gmp: Upgrade formula to v6.2.1

sevan:


mistydemeo:
Thank you! With this and the other dependencies in the GCC 7 tree, are these still compatible with the older versions? Or should I revbump/rebuild the bottle for GCC 7 when I pull them?

sevan:
I didn't try to upgraded existing versions when I put these together, I started with these on the way to building GCC 5.5.
To be on the safe side, I'd revbump/rebuild the compilers.

mistydemeo:
Sounds good. I'll pull these into a staging branch then, and merge all at once.

mistydemeo:
Looks like the compat version of GMP didn't change - it's still `.10`. I'll check the others too when merging.

PR #857, opened 2023-05-30
mpfr: Upgrade formula to v4.2.0

sevan:


Library/Formula/mpfr.rb
```diff
@@ -1,15 +1,9 @@
...
-  url "https://ftp.gnu.org/gnu/mpfr/mpfr-3.1.6.tar.xz"
-  mirror "https://ftpmirror.gnu.org/mpfr/mpfr-3.1.6.tar.xz"
-  sha256 "7a62ac1a04408614fccdc506e4844b10cf0ad2c2b1677097f8f35d3a1344a950"
-
-  bottle do
-    cellar :any
-    sha256 "d40ac2276bc18b7b26170eed452b35707b1dc1ebba5add0244ad5c71c434b3fc" => :tiger_g4e
-    sha256 "194b05ca38b1828ae7c77c3038433459c12249170f76d74af86b71d313b3ec04" => :leopard_g4e
-  end
+  url "https://www.mpfr.org/mpfr-current/mpfr-4.2.0.tar.xz"
```
mistydemeo, line 4: Looks like this URL is dead, but it's available from the GNU FTP:

```suggestion
  url "https://ftp.gnu.org/gnu/mpfr/mpfr-4.2.0.tar.xz"
```

There may be some issues with 4.2.1, which is out now, so we can stick with this version for now. https://github.com/Homebrew/homebrew-core/pull/140157

sevan: ah, ok. Sounds good.

mistydemeo: Thanks!

mistydemeo (approved):

mistydemeo:
Looks like this bumps the mpfr compat version, so I will need to rebuild GCC 7 later.

sevan:
That makes it easier :)

PR #856, opened 2023-05-30
libmpc: Upgrade formula to v1.3.1

sevan:

Issue #152, opened 2013-12-06
gfortran failed to build on 10.5


jrector:
https://gist.github.com/anonymous/7833466

PR #858, opened 2023-05-30
gcc5: Upgrade formula to v5.5

sevan:
[GCC Bug 25127](https://gcc.gnu.org/bugzilla//show_bug.cgi?id=25127) is something to beware of, as discussed in issue  #152.
Affects building on both Tiger & Leopard.
Workaround is to not use CPU optimisation flags.

Built with #855, #856, #857

### DIFF
--- a/Library/Formula/gcc5.rb
+++ b/Library/Formula/gcc5.rb
@@ -21,15 +21,9 @@ class Gcc5 < Formula
 
   desc "The GNU Compiler Collection"
   homepage "https://gcc.gnu.org"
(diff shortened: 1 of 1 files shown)


EricFromCanada:
See also: #571

mistydemeo:
`/usr/local/Cellar/gcc5/5.5.0: 842 files, 168M, built in 553.0 minutes`

I'll merge this and then circle back to @EricFromCanada's PR - sorry I never got around to it before now!

PR #860, opened 2023-05-31
swig: Upgrade formula to v4.1.1

sevan:
Switch to using pcre2 as pcre is EoL

### DIFF
--- a/Library/Formula/swig.rb
+++ b/Library/Formula/swig.rb
@@ -1,19 +1,12 @@
 class Swig < Formula
   desc "Generate scripting interfaces to C/C++ code"
   homepage "http://www.swig.org/"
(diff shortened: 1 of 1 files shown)


mistydemeo (commented):
It looks like `brew test swig` is failing for me; do you know if it's an issue with the built formula, or with the rest being out of date?

sevan:
Issue is with the swig tests case. It tries to make use of ruby.h from Ruby.Framework but that didn't exist until Leopard. We could just the version included in Tigerbrew?

mistydemeo:
Tigerbrew's bundled Ruby won't really work since it's statically linked, but I can look into finding something else for it to work with. Doesn't need to be a blocker on merging this though - sorry for the delay!

PR #810, opened 2023-05-17
openssh: Add formula for v9.3p1

sevan:
Obtained from
https://github.com/Homebrew/homebrew-core/blob/e83e6ef77caf25382a7e9e85c16231361a52eee2/Formula/openssh.rb and modified slightly to drop libfido requirement for now.

Uses updated zlib from tigerbrew so depends on PR https://github.com/mistydemeo/tigerbrew/pull/809

Ideally should be merged after PR https://github.com/mistydemeo/tigerbrew/pull/805

PR #861, opened 2023-05-31
git: Upgrade formula to v2.40.1

sevan:
Git now requires pcre2, and a more recent version of make than is bundled with Tiger at least.
Use curl from tigerbrew regardless of OS version as we need to use modern openssl.
gitk now specifies a different set of colours specified in hex.
Built on Tiger with GCC 4.0 using `--with-brewed-svn --with-pcre2`

Needs #805, #810, #811, #836 and ideally #860

### DIFF
--- a/Library/Formula/git.rb
+++ b/Library/Formula/git.rb
@@ -1,8 +1,8 @@
 class Git < Formula
   desc "Distributed revision control system"
   homepage "https://git-scm.com"
-  url "https://www.kernel.org/pub/software/scm/git/git-2.10.2.tar.xz"
-  sha256 "94802903dd707d85ca3b9a2be35e936a54ce86375f52c6a789efe7ce7e238671"
+  url "https://www.kernel.org/pub/software/scm/git/git-2.40.1.tar.xz"
+  sha256 "4893b8b98eefc9fdc4b0e7ca249e340004faa7804a433d17429e311e1fef21d2"
   head "https://github.com/git/git.git", :shallow => false
 
   bottle do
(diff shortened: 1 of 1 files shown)


Library/Formula/git.rb
```diff
@@ -11,34 +11,32 @@ class Git < Formula
...
   depends_on :expat
-  depends_on "pcre" => :optional
-  depends_on "gettext" => :optional
+  depends_on "pcre2" => :optional
+  depends_on "gettext"
   depends_on "openssl"
-  depends_on "curl" if MacOS.version < :snow_leopard || build.with?("brewed-curl")
+  depends_on "openssh" => :run
+  depends_on "curl"
+  depends_on "make" => :build
```
mistydemeo, line 39: Strangely, I'm seeing gmake fail at the `malloc` stage which is tanking the build for me. I wonder if I need to try building make with a different compiler.

mistydemeo:
```
gmake(27946) malloc: ***  Deallocation of a pointer not malloced: 0x3029bc; This could be a double free(), or
free() called with the middle of an allocated block; Try setting environment variable MallocHelp to see tools
to help debug
gmake(27946) malloc: ***  Deallocation of a pointer not malloced: 0xa00064a0; This could be a double free(), o
r free() called with the middle of an allocated block; Try setting environment variable MallocHelp to see tool
s to help debug
gmake(27946) malloc: *** error for object 0x3029bc: pointer being reallocated was not allocated
gmake(27946) malloc: *** set a breakpoint in szone_error to debug
gmake: *** virtual memory exhausted
```

sevan: hmm, I built gmake 4.4 using the stock GCC 4.0.1 on Tiger and used that.
Is there anything I need to do beyond `brew install -v git` to reproduce your scenario?

mistydemeo: I rebuilt with GCC 4.0 and it works now - I wonder if it was broken with GCC 4.2.

sevan: behaviour is definitely weird. Do you get the same result if you build with GCC 4.2 on Tiger but turn off compiler optimisation in the formula?

mistydemeo (approved):

PR #863, opened 2023-06-07
bitchx: Build fix for Tiger and current OpenSSL

sevan:
Update autoconfig bits to support OpenSSL 1.1.x and newer, as well as support building on Tiger with legacy toolchain.

Needs https://github.com/mistydemeo/tigerbrew/pull/805


mistydemeo (approved):
I wasn't able to get this to build until I added an `ENV.deparallelize` statement. It seems it overwhelmed my system when parallelized.

PR #864, opened 2023-06-10
av98: Add a formula for AV-98 v1.0.1

sevan:
That gives us a Gemini client via Python 3.

Ideally needs https://github.com/mistydemeo/tigerbrew/pull/805,  #823

### DIFF
--- a/Library/Formula/av98.rb
+++ b/Library/Formula/av98.rb
@@ -1,0 +1,24 @@
+class Av98 < Formula
+  desc "AV-98 is an experimental client for the Gemini protocol"
+  homepage "https://tildegit.org/solderpunk/AV-98"
+  url "https://tildegit.org/solderpunk/AV-98/archive/v1.0.1.tar.gz"
+  mirror "https://geeklan.co.uk/files/gemini/av98-1.0.1.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/av98.rb
```diff
@@ -0,0 +1,23 @@
+class Av98 < Formula
+  desc "AV-98 is an experimental client for the Gemini protocol"
+  homepage "https://tildegit.org/solderpunk/AV-98"
+  url "https://tildegit.org/solderpunk/AV-98/archive/v1.0.1.tar.gz"
```
mistydemeo, line 4: Hmm, this is serving a 500. Is there a good mirror somewhere?

sevan: Thanks for the heads up, it seems like the Tilde services have been experiencing issues in the last week, similar problem with another of their services. I'll see if I can find a mirror and get someone to poke at it.

mistydemeo: Looks like it's still down, unfortunately. :/

mistydemeo:
Nice, I see you found a mirror! I'll give this a try.

sevan:
Cool, thanks :)

mistydemeo (approved):
I see it's pure Python with no build dependencies, so I won't bother bottling.

PR #865, opened 2023-06-23
openssl3: Add formula for v3.1.2

sevan:
OpenSSL v1.x.x will soon be EoL, in prep, add a formula for v3.
Though nothing is currently pointed to it as a dependency, it paves the way should someone wish to.

Based on
https://github.com/Homebrew/homebrew-core/blob/bdf52e4896818e57839a8110779b8dad97fe505c/Formula/openssl%403.rb

Currently lacks the ability to support ppc64 because I don't know how to Ruby and handle arch_args, so currently only 32bit builds are supported on ppc.

### DIFF
--- a/Library/Formula/openssl3.rb
+++ b/Library/Formula/openssl3.rb
@@ -1,0 +1,94 @@
+class Openssl3 < Formula
+  desc "Cryptography and SSL/TLS Toolkit"
+  homepage "https://openssl.org/"
+  url "https://www.openssl.org/source/openssl-3.1.1.tar.gz"
+  mirror "https://www.mirrorservice.org/sites/ftp.openssl.org/source/openssl-3.1.1.tar.gz"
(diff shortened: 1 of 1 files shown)


sevan:
This is a lengthy compile btw on a single core G5.

mistydemeo:
> Currently lacks the ability to support ppc64 because I don't know how to Ruby and handle arch_args, so currently only 32bit builds are supported on ppc.

Maybe I can help with that! What's the problem your'e running into?

sevan:
Oh, OpenSSL supports powerpc & powerpc64 builds but when I've submitted covers powerpc, need to fill in the part for 64-bit builds (not even sure it would work, haven't touched 64-bit builds for anything else so far)

mistydemeo (approved):
The build took ages! Confirmed it worked - took about three hours on a G5.

sevan:
🤯

PR #866, opened 2023-06-23
tenfourfox: Upgrade formula to vFPR32.5

sevan:
While here, switch to using SHA2-256 hashes

### DIFF
--- a/Library/Formula/tenfourfox.rb
+++ b/Library/Formula/tenfourfox.rb
@@ -3,24 +3,24 @@ require 'formula'
 class Tenfourfox < Formula
   homepage 'http://www.floodgap.com/software/tenfourfox/'
   urlbase = 'https://downloads.sourceforge.net/tenfourfox/'
-  version '31.0.0'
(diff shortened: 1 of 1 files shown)


Library/Formula/tenfourfox.rb
```diff
@@ -3,24 +3,24 @@
 class Tenfourfox < Formula
   homepage 'http://www.floodgap.com/software/tenfourfox/'
   urlbase = 'https://downloads.sourceforge.net/tenfourfox/'
-  version '31.0.0'
+  version 'fpr32.5'
```
mistydemeo, line 6: Tigerbrew won't recognize this version as newer than 31.0.0; how about we reformat it to `32.5.0-fpr`, so `brew upgrade` works?

sevan: np, $version was used to construct the url to fetch from based on the CPU arch but I've hardcoded it there.

mistydemeo (approved):

PR #871, opened 2023-06-25
cyrus-sasl: Upgrade formula to v2.1.28

sevan:
Build struggles with krb5_gss_register_acceptor_identity definition, skip Kerberos support.
Mutt had no issue sending via SMTP service using TLSv1.3, though not tested with GMail.

Requires https://github.com/mistydemeo/tigerbrew/pull/805


mistydemeo:
Rebased into the openssl branch - thanks!

PR #872, opened 2023-06-25
mutt: Upgrade formula to v2.2.10

sevan:
Project is no longer mercurial based and the dev.mutt.org hostname is dead.
Skip Kerberos support since updated cyrus-sasl formula disables it.
Tested with s-lang support, but not gpgme.

Requires https://github.com/mistydemeo/tigerbrew/pull/805, #871


mistydemeo (approved):
Thanks!

PR #873, opened 2023-06-26
muttils: Upgrade formula to v1.4

sevan:
After v1.4, head dropped Python 2 support and now requires 3.8 min.
Since we currently package 3.7, drop support for building head for now.

### DIFF
--- a/Library/Formula/muttils.rb
+++ b/Library/Formula/muttils.rb
@@ -1,16 +1,9 @@
 class Muttils < Formula
   desc "Provides utilities for use with console mail clients, eg. Mutt."
-  homepage "https://bitbucket.org/blacktrash/muttils/"
-  url "https://bitbucket.org/blacktrash/muttils/get/1.3.tar.gz"
-  sha256 "c8b456b660461441de8927ccff7e9f444894d6550d0777ed7bd160b8f9caddbf"
-  bottle do
-    cellar :any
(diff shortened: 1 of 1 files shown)


Library/Formula/muttils.rb
```diff
@@ -1,16 +1,8 @@
...
-  bottle do
-    cellar :any
-    sha256 "d976f7445a3142ff1c311cf19302b5fcc2976b0a26c0bb48e57e157bd4c7002f" => :yosemite
-    sha256 "01f0c26274540336fa829a8718bb0c3a2a5b5aa3c96c1f4ec7cf79e6263b837b" => :mavericks
-    sha256 "63e819c0bb96a56ed0f159ab816aeac84805a52333bd23298a17cd2abddcb17b" => :mountain_lion
-  end
-
-  head "https://bitbucket.org/blacktrash/muttils", :using => :hg
+  homepage "https://hg.phloxic.productions/muttils"
+  url "https://hg.phloxic.productions/muttils/archive/07e310d44b55.tar.bz2"
```
mistydemeo, line 4: Mind adding a version tag? Otherwise Tigerbrew won't know that this is version 1.4, and it'll make an upgrade to a later numbered version harder.

sevan: Done.

mistydemeo (approved):

PR #881, opened 2023-06-30
zsh-lovers: Upgrade formula to v0.9.1

sevan:


mistydemeo (approved):
Well, that's an easy install.

PR #885, opened 2023-07-13
retroforth: Add formula for v2023.3

sevan:
Need to use a compiler with C11 type redefinition support such as GCC 4.6 or newer.
Tested with GCC 5
`brew install --cc=gcc-5 retroforth`


mistydemeo (changes requested):
Could you add `fails_with :gcc_4_0` and `fails_with :gcc`? That'll prevent Tigerbrew from trying to build this with 4.0 or 4.2.

mistydemeo (approved):
Tested, and confirmed the compiler selection worked! It fell back to GCC 7 for me.

PR #887, opened 2023-07-14
graphviz: Upgrade formula to v2.46.0

sevan:
v2.46.0 is the last version before C++11 support became a requirement. Drop building Quartz support on Tiger as support is lacking. Quartz support is enabled on Leopard and newer due to when Availability.h first landed which the plugin includes and has not been build tested.

### DIFF
--- a/Library/Formula/graphviz.rb
+++ b/Library/Formula/graphviz.rb
@@ -1,23 +1,16 @@
 class Graphviz < Formula
   desc "Graph visualization software from AT&T and Bell Labs"
   homepage "http://graphviz.org/"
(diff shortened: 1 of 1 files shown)


mistydemeo (approved):

mistydemeo:
> v2.46.0 is the last version before C++11 support became a requirement.

I wonder how feasible it is to build newer versions with newer GCCs.

sevan:
I've been trying to build as much as I can with the xcode GCC. Sometimes I get failures when building with GCC 5, and I suspect it may be to do with using the newer compiler from Tigerbrew using the old ld/cctools from xcode.
Updating the other GCC formulas are on my todo list.

PR #889, opened 2023-07-14
vala: Upgrade formula to v0.56.9

sevan:
Need libxslt since the version shipped with OS is likely out of date & the build breaks.

Requires #887, #888


mistydemeo (approved):
Ended up having to revision glib before I could build this due to the libffi bump.

PR #894, opened 2023-07-21
utf8proc: Add formula for v2.8.0

sevan:
Based on
https://raw.githubusercontent.com/Homebrew/homebrew-core/05399d9c3f9ffd0a1424431c34acfe3ab279d477/Formula/utf8proc.rb Test has been changed since strnlen() is not available Simplify test so version doesn't need to be maintained Set standard to C99 to avoid having to return something, since GCC 4.0 defaults to C89 and complains.

Required for new versions of tmux.


mistydemeo:
Looks like this fails with Xcode's ld on Tiger; I'll update it to pull in ld64 as necessary.

mistydemeo (approved):

PR #896, opened 2023-07-21
dovecot: Upgrade formula to v2.3.20

sevan:
Build with PAM support by default, Apple's build of dovecot would patch in OpenDirectory support so auth would likely work without PAM, however without such patch auth fails since user account live in OD, not in the passwd file.

Dovecot detects that kqueue is available and tries to use it for ioloop and notify, however, the version of kqueue in Leopard and Tiger is too old. So while it'll compile ok, it'll fail at runtime when it tries to set the events on startup.

Should be added to the OpenSSL branch I guess, requires #805 

Skipping on making a recommendation for the pam config to add into /etc/pam.d for now since the documentation's advice for OS X doesn't apply to Tiger (they advise using `pam_opendirectory.so` which doesn't exist on the desktop OS. something available on OS X Server?)

### DIFF
--- a/Library/Formula/dovecot.rb
+++ b/Library/Formula/dovecot.rb
@@ -1,24 +1,18 @@
 class Dovecot < Formula
   desc "IMAP/POP3 server"
   homepage "http://dovecot.org/"
-  url "http://dovecot.org/releases/2.2/dovecot-2.2.18.tar.gz"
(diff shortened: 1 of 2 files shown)


mistydemeo (approved):
Merged, thanks!

PR #899, opened 2023-07-23
bzip2: Add a formula for v1.0.8

sevan:
It has received many CVEs over the years.


mistydemeo (approved):
Thank you! Good idea - definitely worth including.

PR #901, opened 2023-07-25
libiconv: Upgrade formula to v1.17

sevan:


Library/Formula/libiconv.rb
```diff
@@ -22,8 +15,6 @@ class Libiconv < Formula
   patch :DATA
 
   def install
-    ENV.deparallelize
```
mistydemeo, line 25: I'm still getting parallel build issues on a G5. Mind putting this back?

mistydemeo (changes requested):

Library/Formula/libiconv.rb
sevan: done.

mistydemeo (approved):
Thank you!

PR #908, opened 2023-07-28
irssi: Upgrade formula to v1.4.4

sevan:
OpenSSL support is required now.
Use up to date dependencies rather than what provided by OS for security.
There is currently a bug with building botti when Perl support is also enabled, so disable it for now.
https://github.com/irssi/irssi/issues/1374

Should go to the OpenSSL branch, needs #805, or should be merged after, with #878

Only tested on Tiger, needs testing on newer OS'.

### DIFF
--- a/Library/Formula/irssi.rb
+++ b/Library/Formula/irssi.rb
@@ -1,10 +1,8 @@
 class Irssi < Formula
   desc "Modular IRC client"
   homepage "http://irssi.org/"
-  url "http://irssi.org/files/irssi-0.8.17.tar.gz"
-  mirror "https://mirrors.kernel.org/debian/pool/main/i/irssi/irssi_0.8.17.orig.tar.gz"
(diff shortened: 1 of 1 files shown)


mistydemeo (approved):
Haven't been able to test on Leopard, but looks good.

Library/Formula/irssi.rb
```diff
@@ -14,43 +12,33 @@ class Irssi < Formula
...
   option "with-dante", "Build with SOCKS support"
   option "without-perl", "Build without perl support"
 
   depends_on "pkg-config" => :build
   depends_on "glib"
-  depends_on "openssl" => :recommended
+  depends_on "openssl"
   depends_on "dante" => :optional
+  depends_on "ncurses"
+  depends_on "perl" => :build if build.with? "perl"
```
mistydemeo, line 23: Does it not require perl at runtime?

sevan: oops, yes, you're right. Let me fix that.

PR #911, opened 2023-07-28
p11-kit: Upgrade formula to v0.24.1

sevan:
Newer versions from 0.25.0 have a strnlen(3) dep which Tiger & Leopard certainly lack.


mistydemeo (commented):
Looks like it can't build with the system libffi on Tiger; mind adding that as a dep?

mistydemeo:
Right, I think I see what happened. libffi isn't keg-only on Tiger - it probably opportunistically picked up the library even though it's not a dep. We definitely need it to build though.

mistydemeo (approved):

PR #913, opened 2023-07-31
clisp build fix

sevan:
Backport fixes from git repo to fix issue with iconv macro and the lack of HTTPS support in clisp. The test suite would previously attempt to connect to the clisp website but since an HTTP->HTTPS redirect is in place now, it would trip up & the test suite run would fail.
Add fix to report an error when attempting to connect to HTTPS urls and fix the test suite to not make network connections.

Tested on Tiger with GCC 4.0.1

### DIFF
--- a/Library/Formula/clisp.rb
+++ b/Library/Formula/clisp.rb
@@ -4,7 +4,7 @@ class Clisp < Formula
   url "http://ftpmirror.gnu.org/clisp/release/2.49/clisp-2.49.tar.bz2"
   mirror "https://ftp.gnu.org/gnu/clisp/release/2.49/clisp-2.49.tar.bz2"
   sha256 "8132ff353afaa70e6b19367a25ae3d5a43627279c25647c220641fed00f8e890"
-  revision 1
+  revision 2
 
   bottle do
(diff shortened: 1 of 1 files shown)


Library/Formula/clisp.rb
```diff
@@ -29,7 +39,7 @@ class Clisp < Formula
 
   def install
     ENV.deparallelize # This build isn't parallel safe.
-    ENV.O0 # Any optimization breaks the build
+    ENV.no_optimization # Build itself sets optimisation flags.
```
mistydemeo, line 42: If that's true, is it safe for us to just remove this entirely? Otherwise we might override it. `no_optimization` is an alias for `O0`.

sevan: No, because if you don't state anything, -mcpu & -Os will be specified, right?

mistydemeo: `no_optimization` doesn't strip all optimization flags from the CFLAGS; it sets `O0` explicitly.

mistydemeo: No, actually, I'm totally wrong. For some reason I remembered otherwise!

mistydemeo (approved):

PR #914, opened 2023-07-31
gambit-scheme: Upgrade formula to v4.9.5

sevan:
Make building with --enable-single-host optional since that requires a newer compiler.
Enable the dynamic-clib optimisation since the build suggests it.
By setting names, we avoid the conflict with Ghostscript. Add an option to build with OpenSSL support.
Turned off optimisation since the build process sets different optimisation flags for different components varying from O1 to O3. The build process supports more options related to compiler optimisation flags, but I did not explore any further.

Tested with GCC 4.0.1 on Tiger with OpenSSL 1.1.1u
OpenSSL support is optional, but it requires v1.1.1, so needs to be added to the OpenSSL branch for #805 or merged after if it is going to built it.

When running 'make check' there's 28 tests which fail since the unit tests try to load files from the install destination directory. [gambit/#814](https://github.com/gambit/gambit/issues/814)

### DIFF
--- a/Library/Formula/gambit-scheme.rb
+++ b/Library/Formula/gambit-scheme.rb
@@ -1,40 +1,37 @@
 class GambitScheme < Formula
   desc "Complete, portable implementation of Scheme"
-  homepage "http://dynamo.iro.umontreal.ca/~gambit/wiki/index.php/Main_Page"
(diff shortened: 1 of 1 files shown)


mistydemeo (approved):
Took a solid hour to build on my G5; glad it works!

PR #915, opened 2023-08-01
poke: Add formula for v3.2

sevan:
Formula adapted from
https://raw.githubusercontent.com/Homebrew/homebrew-core/f357e4c65c5a923db687100ac4167232299b3983/Formula/poke.rb

Tested on Tiger with GCC 4.0.1


mistydemeo (approved):
Thank you! It failed for me with GCC 4.2 but worked iwth 4.0, so I'll tag the compatibility appropriately.

PR #917, opened 2023-08-08
lighttpd 1.4.71

sevan:
Drop LDAP support since OpenLDAP in Leopard & Tiger lack support for ldap_distroy() which the module calls.
Provide the missing parameters for getxattr() to fix build. Provide the definition of STAILQ_FOREACH if not found in <sys/queue.h> as with Tiger.
use-ipv6 is nolonger a configurable option.

Requires #805 so should be added to the OpenSSL branch or merged after.

### DIFF
--- a/Library/Formula/lighttpd.rb
+++ b/Library/Formula/lighttpd.rb
@@ -1,24 +1,15 @@
 class Lighttpd < Formula
   desc "Small memory footprint, flexible web-server"
   homepage "http://www.lighttpd.net/"
-  url "http://download.lighttpd.net/lighttpd/releases-1.4.x/lighttpd-1.4.37.tar.xz"
-  sha256 "804e6f60567ca1da2b3927f92b7e9332b93aca9560f282ca135b86b7558979bd"
-
-  bottle do
-    sha256 "5ccc450f4e552726c108b00955c835c8b4c82ae844eb944f588cedc590f3ea43" => :el_capitan
-    sha256 "31ad0f53176a0509c51b52082868a847751be055b2d0c11f16974cdb18bdfa51" => :yosemite
-    sha256 "c2900f5ea5bc9cc6ebdf7e2e86724f21b2f4fed923912f2904f5466b9a040109" => :mavericks
-    sha256 "0b4fd109bba1b5ad7c2ff7299fe495fe646f8bd767ddea78e6fccefeeb095347" => :mountain_lion
-  end
+  url "https://download.lighttpd.net/lighttpd/releases-1.4.x/lighttpd-1.4.71.tar.xz"
+  sha256 "b8b6915da20396fdc354df3324d5e440169b2e5ea7859e3a775213841325afac"
 
   option "with-lua51", "Include Lua scripting support for mod_magnet"
 
   depends_on "pkg-config" => :build
-  depends_on "autoconf" => :build
-  depends_on "automake" => :build
-  depends_on "libtool" => :build
-  depends_on "pcre"
+  depends_on "pcre2"
   depends_on "openssl"
+  depends_on "xxhash"
   depends_on "lua51" => :optional
   depends_on "libev" => :optional
 
@@ -48,18 +39,15 @@ class Lighttpd < Formula
       --prefix=#{prefix}
       --sbindir=#{bin}
       --with-openssl
-      --with-ldap
       --with-zlib
       --with-bzip2
       --with-attr
+      --with-xxhash
     ]
 
     args << "--with-lua" if build.with? "lua51"
     args << "--with-libev" if build.with? "libev"
 
-    # autogen must be run, otherwise prebuilt configure may complain
-    # about a version mismatch between included automake and Homebrew's
-    system "./autogen.sh"
     system "./configure", *args
     system "make", "install"
 
@@ -74,11 +62,6 @@ class Lighttpd < Formula
         s.sub!(/^var\.conf_dir\s*=\s*".+"$/, "var.conf_dir    = \"#{config_path}\"")
         s.sub!(/^server\.port\s*=\s*80$/, "server.port = 8080")
         s.sub!(%r{^server\.document-root\s*=\s*server_root \+ "\/htdocs"$}, "server.document-root = server_root")
-
-        # get rid of "warning: please use server.use-ipv6 only for hostnames, not
-        # without server.bind / empty address; your config will break if the kernel
-        # default for IPV6_V6ONLY changes"
-        s.sub!(/^server.use-ipv6\s*=\s*"enable"$/, 'server.use-ipv6 = "disable"')
 
         s.sub!(/^server\.username\s*=\s*".+"$/, 'server.username  = "_www"')
         s.sub!(/^server\.groupname\s*=\s*".+"$/, 'server.groupname = "_www"')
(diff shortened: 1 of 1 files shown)


gstrauss:
Hi @sevan  Thanks for putting this together.  I am a lighttpd developer and will accept (reasonable) portability patches upstream, such as preprocessor options like
```
#if defined(__APPLE__) && defined(__MACH__)
#include <TargetConditionals.h> /* TARGET_OS_IPHONE, TARGET_OS_MAC */
...
#endif
```
in order to use a different function signature for `getxattr()` where needed for older MacOS, though I do not have access to such systems on which to test, and am not up-to-speed on digging through Apple developer docs for this historical information.  As an option, you can also build lighttpd *without* `--with-attr`, as I do not believe that xattr is widely used with web servers.

Is there a reason the openldap libraries on Leopard and Tiger are too old to have `ldap_destroy()`?  Could openldap be updated?  Of course I agree that the quick answer is to disable the optional LDAP feature in lighttpd since as you say, support for `ldap_destroy()` does not exist on Tiger.  The change in lighttpd to use `ldap_destroy()` was added for https://redmine.lighttpd.net/issues/2849  A commenter noted that `ldap_destroy()` is not available before openldap 2.4 (released Oct 2007 !), but I have not verified that.  If there is something appropriate to use instead of `ldap_destroy()` for the old openldap on Tiger, please let me know.

You may wish to submit your STAILQ_FOREACH patch upstream to https://github.com/litespeedtech/ls-hpack

FYI: openssl 1.1.1 will reach end-of-life this September: https://www.openssl.org/blog/blog/2023/03/28/1.1.1-EOL/

sevan:
Hi @gstrauss
I suspect my patch for `getxattr()` to add the missing parameters applies to more recent versions on macOS, since the signature has not changed between 10.4 where the function was introduced and 13.3.

[This is the prototype](https://github.com/apple-oss-distributions/xnu/blob/aca3beaa3dfbd42498b42c5e5ce20a938e6554e5/bsd/sys/xattr.h#L88) in `xattr.h` for the latest macOS source release available (13.3).
On OS X 10.4 Tiger it's
`ssize_t getxattr(const char *path, const char *name, void *value, size_t size, u_int32_t position, int options);`
I haven't tried to build the latest version on a more recent macOS release `--with-attr` to confirm though.

Regarding OpenLDAP on Tiger, the OS itself was released in 2005 
```
% /usr/libexec/slapd -VV
@(#) $OpenLDAP: slapd 2.2.19 $
```
Since it wasn't packaged already I refrained from packaging it but I guess we could package a new version of OpenLDAP and use that if it is buildable on Tiger.

OpenSSL 3 is packaged (#865), no idea how it'll be handled within Tigerbrew in terms of integration.

sevan:
[Raised a pull request for the ls-hpack change](https://github.com/litespeedtech/ls-hpack/pull/13)

sevan:
I have of course made the mistake of assuming that `getxattr()` is a feature in Darwin. Looks like the codebase assumes the Linux implementation and the [parameters match that signature](https://man7.org/linux/man-pages/man2/getxattr.2.html).
```
ssize_t getxattr(const char *path, const char *name, void value[.size], size_t size);
```

I'll raise the relevant changes upstream as you suggested.

gstrauss:
@sevan, why does your patch remove `system "./autogen.sh"` from the build steps?

FYI: lighttpd no longer provides an option to use libev, so that option should probably be removed from the config file.

For `s.sub!(/^server\.event-handler\s*=\s*"linux-sysepoll"$/, 'server.event-handler = "select"')`, "kqueue" should be preferred, if available on Tiger (and only if there were not major bugs with kqueue on MacOS way back in Tiger), or else "poll" should be preferred to "select".  If you comment out the `server.event-handler` line in lighttpd.conf, lighttpd will be default use kqueue, if available, or else poll, if available, and finally select if none of the others are available.

I've added the following patch for the next release of lighttpd.  Compile-tested on MacOS 11 (Big Sur):
```
--- a/src/stat_cache.c
+++ b/src/stat_cache.c
@@ -841,8 +841,13 @@ static int stat_cache_attr_get(const char *name) {
   #if defined(HAVE_XATTR)
    #if defined(HAVE_SYS_XATTR_H)
     ssize_t attrlen;
+   #if defined(__APPLE__) && defined(__MACH__)
+    if (0 < (attrlen = getxattr(name, attrname,
+                                attrval, sizeof(attrval)-1, 0, 0)))
+   #else
     if (0 < (attrlen = getxattr(name, attrname,
                                 attrval, sizeof(attrval)-1)))
+   #endif
    #else
     int attrlen = sizeof(attrval)-1;
     if (0 == attr_get(name, attrname, attrval, &attrlen, 0))
```
However, unless there is a need, you probably should compile lighttpd *without* `--with-attr`.  Even with the support compiled in, lighttpd does waste effort trying to use xattr for Content-Type unless enabled in lighttpd.conf with `mimetype.use-xattr = "enabled"`

sevan:
I dropped the call to autogen because the stock release source archive we build from is autotoolsed up and since we don't touch any of that infra, we don't need to re-run again. The configure stage also didn't complain about automake either so it seemed safe to drop & it saves on a lenghty build process to get those things installed before you get to build lighttpd if you don't have prebuilt packages at hand.

Just a thought regarding kqueue since I haven't tried building with it, though Tiger & Leopard have kqueue, it is not so much that it's buggy but it is very old and missing functionality in newer versions so like the case for Dovecot, which uses ioloop & notify from kqueue but it only checks if kqueue is at all available, build passes but when it comes to run time, it fails. Are you explicitly checking for the functionality from kqueue which you require or just that kqueue is available?

I'll drop the attr & libev bits in the formula.

gstrauss:
It if works without the `./autogen.sh` step, then that's fine.   The comment removed in your patch gave me some concern:
```
-    # autogen must be run, otherwise prebuilt configure may complain
-    # about a version mismatch between included automake and Homebrew's
-    system "./autogen.sh"
```

lighttpd does not explicitly check for functionality from `kqueue`(), but only uses basic `kqueue()` functionality by default: `EVFILT_READ` or `EVFILT_WRITE` with `EV_ADD` or `EV_DELETE`.  lighttpd can be configured (`server.stat-cache-engine = "kqueue"`)  to use `kqueue()` for `EVFILT_VNODE`, but that is not done by default and `EVFILT_VNODE` and flags are defined in the Apple developer man page for `kqueue()` dated Apr 14, 2000.  I do not have access to ancient Mac OS Tiger systems on which to test, so can not test the `kqueue()` functionality, and I also can not test the `sendfile()` functionality, so I leave those to your discretion.

gstrauss:
I'd recommend removing `--with-bzip2` from the build, too.  In practice, that is not widely used, if at all.  brotli is much more popular.

sevan:
> lighttpd does not explicitly check for functionality from `kqueue()`, but only uses basic `kqueue()` functionality by default: `EVFILT_READ` or `EVFILT_WRITE` with `EV_ADD` or `EV_DELETE`. lighttpd can be configured (`server.stat-cache-engine = "kqueue"`) to use `kqueue()` for `EVFILT_VNODE`, but that is not done by default and `EVFILT_VNODE` and flags are defined in the Apple developer man page for `kqueue()` dated Apr 14, 2000. I do not have access to ancient Mac OS Tiger systems on which to test, so can not test the `kqueue()` functionality, and I also can not test the `sendfile()` functionality, so I leave those to your discretion.

All those are available in `sys/event.h` so I enabled it. Skipping on sendfile as it is technically there but guarded off (officially it was introduced in 10.5).

gstrauss:
@sevan your changes look good to me.

With the removal of `--with-attr`, you probably should *remove* your patch to stat_cache.c
```
-                                attrval, sizeof(attrval)-1)))
+                                attrval, sizeof(attrval)-1, 0, NULL)))
```
or else that will conflict with future releases of lighttpd (which will contain the patch I posted further above)

FYI: if the input file is lighttpd source `doc/config/lighttpd.conf`, then you do not have to uncomment `server.event-handler`, as that will use the best OS event framework available, i.e. kqueue on Mac OS.

FYI: If you're not sure about `sendfile()` support, it is still a good idea to specify `server.network-backend = "writev"` since if not specified, and presence of `sendfile()` is detected, then `sendfile()` will be used by default by lighttpd to send files.  However, if `sendfile()` fails, lighttpd will transparently fallback to `writev()`, so if `sendfile()` always fails, it will waste a syscall each time sending files to rediscover `sendfile()` will fail.

sevan:
Decided to leave the patch there should someone want to build with attr support. It'll be dropped when the formula is upgraded in the future as the patch won't apply.

mistydemeo:
Thank you so much everyone for all the work on this. Just to check, is this PR now ready for merge/binary builds?

gstrauss:
LGTM.  Thanks!

sevan:
Yep, ready to land.

mistydemeo (approved):
Thank you!

PR #919, opened 2023-08-11
Library/Homebrew/software_spec.rb: Save on the 301 redirect

sevan:
Old $HOMEBREW_BOTTLE_DOMAIN redirects, so save on the roundtrip every time.


mistydemeo:
Sorry, closed this by accident.

mistydemeo (approved):
Thanks!

PR #920, opened 2023-08-11
curl 8.2.1

sevan:
Drop libidn support since curl requires libidn2 which is currently not packages and the option is no longer recognised by configure script. Rework so we only ever use openssl from tigerbrew, it's unlikely we'll ever want to use the version supplied by OS on legacy release now.


mistydemeo (approved):
It looks like there are a handful of build changes from upstream; I'm going to look at incorporating them. https://github.com/Homebrew/homebrew-core/blob/master/Formula/curl.rb#L78-L79

Issue #927, opened 2023-08-12
Cannot install Perl


Amatarou-SUN:
I am getting an error when trying to install perl.
How do I install it?
![Picture 1](https://github.com/mistydemeo/tigerbrew/assets/85050558/1943a0a7-a017-4a87-bcd3-a87fd32657a6)

mistydemeo:
Can you please post the output of running `brew --config`?

Amatarou-SUN:
Sure!
![Picture 2](https://github.com/mistydemeo/tigerbrew/assets/85050558/09f78c48-5218-4b86-9bec-5c5bf2cc14e7)

mistydemeo:
Thank you! Can you also provide me the logs located in `~/Library/Logs/Homebrew/perl`? You can upload them to gist.github.org

Amatarou-SUN:
https://gist.github.com/Amatarou-SUN/542f785ed617db6e1369d9df6b21c9a2

mistydemeo:
I'm surprised it's setting `MACOSX_DEPLOYMENT_TARGET`. Do you have that set in your environment?

Amatarou-SUN:
How to check it?

Amatarou-SUN:
how do i check? I've googled but can't find a way

Amatarou-SUN:
I found how, but MACOSX_DEPLOYMENT_TARGET is empty.

PR #928, opened 2023-08-12
Fix perl build on 10.5

sevan:
Deployment target needs to be set to 10.5 if -rpath is used during build.

Resolves issue #927


mistydemeo (approved):

mistydemeo:
Thank you! Great find.

Amatarou-SUN:
Installing perl worked! thank you everyone

PR #930, opened 2023-08-12
git 2.41.0

sevan:
v2.40.1 formula was missing support for Leopard in `config.mak.uname` patch.
Tested on Tiger with GCC 4.0.1 & Leopard with GCC 4.2.

`git-credential-osxkeychain` is only build tested, I don't have a working setup.


mistydemeo (approved):
Thank you! I haven't been able to test the keychain yet either, but let's go ahead and merge.

PR #933, opened 2023-08-13
qemu 2.3.1

sevan:
Previously, this formula was reverted from 2.4.x to 2.2.1 due to lockups experienced on 10.5.8 on G5. 2.3.1 runs ok for me on a G5 running Tiger.
Drop gnutls as it is dangerously out of date and disable VNC TLS support which requires it.
Allow the guest agent to be built in case OS X is virtualised.
The user mode is not enabled by default, there's no need to disable it.
Update the url for the test image with a working URL.

Built on Tiger with GCC 5
i386 hypervisor tested with
`qemu-system-i386 -cdrom some-bootable.iso`
PowerPC hypervisor tested with
`qemu-system-ppc some.iso`
Then connected to the console via VNC on localhost using [Chicken](https://sourceforge.net/projects/chicken/).

Needed the fix from PR #765

### DIFF
--- a/Library/Formula/qemu.rb
+++ b/Library/Formula/qemu.rb
@@ -1,19 +1,15 @@
 class Qemu < Formula
   desc "x86 and PowerPC Emulator"
   homepage "http://wiki.qemu.org"
-  # 2.2.1 is the last version to work pre-10.6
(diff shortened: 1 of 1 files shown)


mistydemeo (approved):
Thank you!

Issue #430, opened 2016-08-27
tcl-tk 8.6.6 breaks existing Tiger patch


dcwar:
Hi! Looks like they're now including a new version of sqlite (3.13.0), which breaks the existing Tiger patch for tcl-tk. Build fails as a result.

programmingkidx:
The patched file is now located here: tcl8.6.6/pkgs/sqlite3.13.0/compat/sqlite3/sqlite3.c
Changing the file that is patched shouldn't be difficult.

programmingkidx:
Please replace the old patch in the rb file with this one:

diff --git a/pkgs/sqlite3.13.0/compat/sqlite3/sqlite3.c b/pkgs/sqlite3.13.0/compat/sqlite3/sqlite3.c
index 3e20b56..95035d4 100644
--- a/pkgs/sqlite3.13.0/compat/sqlite3/sqlite3.c
+++ b/pkgs/sqlite3.13.0/compat/sqlite3/sqlite3.c
@@ -20140,6 +20140,7 @@ SQLITE_PRIVATE void sqlite3MemSetDefault(void){
 #include <sys/sysctl.h>
 #include <malloc/malloc.h>
 #include <libkern/OSAtomic.h>
+
 static malloc_zone_t* _sqliteZone_;
 #define SQLITE_MALLOC(x) malloc_zone_malloc(_sqliteZone_, (x))
 #define SQLITE_FREE(x) malloc_zone_free(_sqliteZone_, (x));
@@ -20147,6 +20148,29 @@ static malloc_zone_t* _sqliteZone_;
 #define SQLITE_MALLOCSIZE(x) \
         (_sqliteZone_ ? _sqliteZone_->size(_sqliteZone_,x) : malloc_size(x))
 
+/*
+** If compiling for Mac OS X 10.4, the OSAtomicCompareAndSwapPtrBarrier
+** function will not be available, but individual 32-bit and 64-bit
+** versions will.
+*/
+
+#ifdef __MAC_OS_X_MIN_REQUIRED
+# include <AvailabilityMacros.h>
+#elif defined(__IPHONE_OS_MIN_REQUIRED)
+# include <Availability.h>
+#endif
+
+typedef int fc_atomic_int_t;
+#if (MAC_OS_X_VERSION_MIN_REQUIRED > MAC_OS_X_VERSION_10_4 || __IPHONE_VERSION_MIN_REQUIRED >= 20100)
[...]

PR #937, opened 2023-08-18
sqlite 3.42.0

sevan:
Tested on
Leopard i386 & ppc, GCC 4.2
Tiger ppc, GCC 4.0.1

PR #938, opened 2023-08-18
tcl-tk 8.6.13

sevan:
Don't build a private copy of SQLite, use the version in Tigerbrew.
TK needs Snow Leopard at least to support Aqua, build fails due to the lack of NSMenuDelegate, NSApplicationDelegate, NSOpenSavePanelDelegate.

Fixes issue #430
Requires PR #937

Tested on
Leopard i386 & ppc, GCC 4.2
Tiger ppc, GCC 4.0.1


mistydemeo (approved):
Thank you!

If anyone requests we can reenable X on 10.6+, but this seems like the right approach for now.

PR #942, opened 2023-08-23
Fix zlib URL to point to fossils/ folder

lazd:
This PR corrects the URL for zlib source code. The URL without the fossils/ folder only works for the latest release.


mistydemeo (approved):
Thank you!

PR #951, opened 2023-09-03
openconnect 9.12

sevan:
Use latest vpnc-script
Switch back to using OpenSSL now that it is up to date (gnutls isn't and it requires C11 support).
Fix test

Build tested on Tiger powerpc with GCC 4.0.1, haven't tried to establish a VPN connection to a supported appliance.

### DIFF
--- a/Library/Formula/openconnect.rb
+++ b/Library/Formula/openconnect.rb
@@ -1,13 +1,10 @@
 class Openconnect < Formula
   desc "Open client for Cisco AnyConnect VPN"
   homepage "http://www.infradead.org/openconnect.html"
-  url "ftp://ftp.infradead.org/pub/openconnect/openconnect-7.06.tar.gz"
-  sha256 "facf695368dc4537a6a30e2147be90b1d77ee3cb2d269eaef070b6d9ddab70f2"
+  url "https://www.infradead.org/openconnect/download/openconnect-9.12.tar.gz"
+  sha256 "a2bedce3aa4dfe75e36e407e48e8e8bc91d46def5335ac9564fbf91bd4b2413e"
 
   bottle do
-    sha256 "7d514ce407dee1972abf23a51ff2acfe7d72bfde9854639bd6097bb39e6ab058" => :yosemite
-    sha256 "1d83fa61da006086115e7d0b2e4ce04577c3812b10e553aa9b1b33b7d450f1de" => :mavericks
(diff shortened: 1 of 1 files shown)


Library/Formula/openconnect.rb
```diff
@@ -17,19 +14,18 @@ class Openconnect < Formula
...
+  depends_on "p11-kit"
   depends_on "stoken" => :optional
   depends_on "libxml2"
+  depends_on "zlib"
 
   resource "vpnc-script" do
-    url "http://git.infradead.org/users/dwmw2/vpnc-scripts.git/blob_plain/a64e23b1b6602095f73c4ff7fdb34cccf7149fd5:/vpnc-script"
-    sha256 "cc30b74788ca76928f23cc7bc6532425df8ea3701ace1454d38174ca87d4b9c5"
+    url "https://git.infradead.org/users/dwmw2/vpnc-scripts.git/blob_plain/78050150f712d96f81dcade1efe5d16d017e8471:/vpnc-script"
+    sha256 "48cdff979f5f941e2740193c4ee2a90f55549908afcb1fcee64a66f8c0fec05f"
```
mistydemeo, line 28: I wonder if the contents of this are unstable? I'm seeing a `sha256` of `cc30b74788ca76928f23cc7bc6532425df8ea3701ace1454d38174ca87d4b9c5`. Maybe worth mirroring to a gist.

When I make suggestions like these, would you rather I make them by pushing to this PR, or holding off until you have the time to look at it?

sevan: I hit the issue with the hash too, the `cc30b74788ca76928f23cc7bc6532425df8ea3701ace1454d38174ca87d4b9c5` hash is for the old version of the script which lingers in your ~/Library/Caches/Homebrew, if you remove and re-attempt the build, everything should fall into place. The new url should be stable. I picked the link to the most recent commit, not the floating version in HEAD which would change per commit.

Regarding feedback, I don't mind.

mistydemeo: Oh of course, good catch! I'll clear it out from the cache before the next build.

mistydemeo (approved):
I had to unlink gnutls to get it to build; wonder if it's worth explicitly turning it off to avoid that. I'll bottle the working build for Tiger so hopefully many users don't end up running into it.

sevan:
Hmm, there is a `--without-gnutls` flag supported by configure.

Issue #965, opened 2023-09-13
zlib never updating and affecting other software


and3r00:
I'm unable to install zlib, git, openssh, and many other things for some reason. OSX Leopard 10.5.9

I installed the linked Xcode, tried to install the newer XQuartz, supposedly it did install, but it continues to show it as not installed.

I edited /etc/paths and moved /usr/local/bin to the top. I also added /usr/local/bin to my path.

https://gist.github.com/and3r00/2cb10c4d06c16be4393dd89982c98292

mistydemeo:
Can you try running `brew install zlib` first? It looks like it's getting confused about which zlib to use; it may help to get that installed first.

mistydemeo:
Can you confirm the output for `brew info openssh`? I want to make sure you're using a version where zlib is a dependency.

and3r00:
ajis-ibook-g4:~ test1$ brew info openssh
openssh: stable 9.4p1
OpenBSD freely-licensed SSH connectivity tools
https://www.openssh.com/
Not installed
From: /usr/local/Library/Formula/openssh.rb
==> Dependencies
Build: pkg-config ✔
Required: ldns ✔, openssl ✔, zlib ✔

sevan:
What does `brew info zlib` say?

and3r00:
ajis-ibook-g4:~ test1$ brew info zlib
zlib: stable 1.3
General-purpose lossless data-compression library
http://www.zlib.net/

This formula is keg-only.
OS X already provides this software and installing another version in
parallel can cause all kinds of trouble.

/usr/local/Cellar/zlib/1.3 (10 files, 424K)
  Built from source
From: /usr/local/Library/Formula/zlib.rb
==> Caveats
This formula is keg-only, which means it was not symlinked into /usr/local.

OS X already provides this software and installing another version in
parallel can cause all kinds of trouble.

Generally there are no consequences of this for you. If you build your
own software and it requires this formula, you'll need to add to your
build variables:

    LDFLAGS:  -L/usr/local/opt/zlib/lib
    CPPFLAGS: -I/usr/local/opt/zlib/include

and3r00:
I'm also getting this error on many different installs. I am not sure if it is unrelated.
```
brew install readline
==> Downloading https://ftpmirror.gnu.org/readline/readline-8.2.tar.gz
Already downloaded: /Users/test1/Library/Caches/Homebrew/readline-8.2.tar.gz
[...]

PR #966, opened 2023-09-13
openssh Fix zlib version check for >= 1.3

sevan:
Not bumping package revision as it only affects new build attempts, post zlib 1.3 install.
Resolves issue #965


mistydemeo (approved):
Thank you!

PR #967, opened 2023-09-14
coreutils 9.2

sevan:
9.3 removed the option for disabling > 2038 date support requirement
9.4 pulls in newer gnulib which doesn't build on Tiger.

Built on Tiger powerpc with GCC 4.0.1.

### DIFF
--- a/Library/Formula/coreutils.rb
+++ b/Library/Formula/coreutils.rb
@@ -1,15 +1,11 @@
 class Coreutils < Formula
   desc "GNU File, Shell, and Text utilities"
   homepage "https://www.gnu.org/software/coreutils"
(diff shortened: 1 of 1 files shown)


mistydemeo:
I can't remember, did you document what the gnulib issue was?

mistydemeo:
Looks like `brew test` is failing here - and I'm able to confirm it manually. `gln`'s behaviour seems off.

sevan:
I'll take a look at the faililng test. gnulib needs to be [built against libsigsegv on Tiger now](https://lists.gnu.org/archive/html/bug-gnulib/2021-08/msg00132.html)

sevan:
Rabbit hole, breakages again indirectly due to gnulib which is why the test target in the formula fails. I found reports but there was no fix. Need to find when the issue was introduced in gnulib.

PR #970, opened 2023-09-17
fennel 1.3.1

sevan:
Based on
https://github.com/Homebrew/homebrew-core/blob/84521a58ec3789c7e1268e050d999bc72b428646/Formula/f/fennel.rb

Requires #969


mistydemeo (approved):
Confirmed working with that other branch merged.

PR #971, opened 2023-09-17
openssl x86 & Snow Leopard fixes

sevan:
Skipping version bump as they would not have built on x86 in the first place.


mistydemeo (approved):
I'm unable to test, but I trust these work.

PR #972, opened 2023-09-17
vim 9.0.1907

sevan:
NSSoundDelegate apparently was introduced in Snow Leopard yet the build breaks on it. Guarding off enabling sound support to El Capitan as that's the next version I had access to for testing. https://developer.apple.com/documentation/appkit/nssounddelegate

Tested on Tiger powerpc with GCC 4.0.1, Snow Leopard with GCC 4.2, El Capitan with llvm from XCode 7.


mistydemeo (approved):
Thank you! Good catch on the sound feature.

PR #974, opened 2023-10-04
openssh 9.5p1

sevan:
Tested on 10.4 (G4) & 10.5 (G5)

Should we think about switch over to OpenSSL 3?


mistydemeo:
> Should we think about switch over to OpenSSL 3?

Definitely seems like a good idea. 👍

mistydemeo (approved):

PR #981, opened 2023-10-24
openssl3 3.1.4

sevan:
Tested on Tiger powerpc, took just over 2 hours.

PR #982, opened 2023-10-25
python3: 3.7.17

sevan:
While here, switch to using openssl3.

Tested on Tiger powerpc
Ideally should be merged after the openssl3 update in #981


mistydemeo (approved):
Looks like some of the test data breaks relocation; I'll push a patch after I merge that fixes that.

PR #983, opened 2023-10-25
openssl3 switch

sevan:
Start switching formulas to openssl3 now that openssl 1.1.1 is EoL.
Ideally should be merged after the openssl3 update in #981

Libraries such as ldns, libevent, libssh2 should be built first, before applications, otherwise openssl 1.1.1 creeps in as a dependency.

### DIFF
--- a/Library/Formula/curl.rb
+++ b/Library/Formula/curl.rb
@@ -3,10 +3,10 @@ class Curl < Formula
   homepage "https://curl.haxx.se/"
   url "https://curl.se/download/curl-8.4.0.tar.xz"
   sha256 "16c62a9c4af0f703d28bda6d7bbf37ba47055ad3414d70dec63e2e6336f2a82d"
+  revision 1
 
   bottle do
     cellar :any
-    sha256 "0065f0535323140e66b9338dc0bd8397db65e6edecbf1f2f456962fef95d04c4" => :tiger_altivec
   end
 
   keg_only :provided_by_osx
@@ -25,7 +25,7 @@ class Curl < Formula
   depends_on "zlib"
 
   if (build.without?("libressl"))
-    depends_on "openssl"
+    depends_on "openssl3"
   end
 
   depends_on "pkg-config" => :build
@@ -52,9 +52,9 @@ class Curl < Formula
       args << "--with-ssl=#{Formula["libressl"].opt_prefix}"
       args << "--with-ca-bundle=#{etc}/libressl/cert.pem"
     else
-      ENV.prepend_path "PKG_CONFIG_PATH", "#{Formula["openssl"].opt_lib}/pkgconfig"
-      args << "--with-ssl=#{Formula["openssl"].opt_prefix}"
-      args << "--with-ca-bundle=#{etc}/openssl/cert.pem"
+      ENV.prepend_path "PKG_CONFIG_PATH", "#{Formula["openssl3"].opt_lib}/pkgconfig"
+      args << "--with-ssl=#{Formula["openssl3"].opt_prefix}"
+      args << "--with-ca-bundle=#{etc}/openssl@3/cert.pem"
     end
 
     args << (build.with?("libssh2") ? "--with-libssh2" : "--without-libssh2")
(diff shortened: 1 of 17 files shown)


sevan:
Libraries such as ldns, libevent, libssh2 should be built first, before applications, otherwise openssl 1.1.1 creeps in as a dependency.

mistydemeo:
Is it okay if I push commits adding bottles to this branch as I work through it? Or would you prefer I work with another branch for that?

sevan:
Sure, I can just pull on the branch with your changes.

mistydemeo:
I've tested everything in this PR and added bottles; looking good so far! Thank you again so much for your work on this.

sevan:
That's great! you're welcome :)
Do you want to merge this lot and I continue with a new separate batch? (I don't mind either way)

mistydemeo:
Sounds great! Mark it ready for review whenever you want.

Library/Formula/curl.rb
```diff
@@ -52,9 +53,9 @@ def install
       args << "--with-ssl=#{Formula["libressl"].opt_prefix}"
       args << "--with-ca-bundle=#{etc}/libressl/cert.pem"
     else
-      ENV.prepend_path "PKG_CONFIG_PATH", "#{Formula["openssl"].opt_lib}/pkgconfig"
-      args << "--with-ssl=#{Formula["openssl"].opt_prefix}"
-      args << "--with-ca-bundle=#{etc}/openssl/cert.pem"
+      ENV.prepend_path "PKG_CONFIG_PATH", "#{Formula["openssl3"].opt_lib}/pkgconfig"
+      args << "--with-ssl=#{Formula["openssl3"].opt_prefix}"
+      args << "--with-ca-bundle=#{etc}/openssl@3/cert.pem"
```
mistydemeo, line 58: Wait a sec, I think this is wrong.

mistydemeo (changes requested):
Caught one minor mistake, but otherwise looks good!

Library/Formula/curl.rb
sevan: Do you mean the path set `--with-ca-bundle`? because the [openssl3 formula sets it to that](https://github.com/mistydemeo/tigerbrew/blob/ee4014d1451a1ae45811e433eb91a27877b63c4d/Library/Formula/openssl3.rb#L63) I thought it was best to keep things separate from the openssl 1.1.1 formula as I suspect things will live side by side for some time.

mistydemeo:
Hmmm. I think that's actually a typo in the OpenSSL 3 formula then - probably when it got adapted from a version named with the `@` upstream. It looks like this means the `etc` path didn't make it into the OpenSSL 3 bottle; I'll look at updating/revision bumping it today, then updating the formula(s) in this PR that referenced that path.

sevan:
So you mean that while the build of the formula will create a directory in `etc` called `openssl@3`, when you bottle it up, it'll be skipped?
I left the `openssl@3` directory reference in so that the two openssl formulas would not overlap.

mistydemeo:
OK, I think I'm wrong about the name being an issue. We do a direct file comparison and grab anything the formula installed in `/etc`:

```ruby
  def build_bottle_preinstall
    @etc_var_glob ||= "#{HOMEBREW_PREFIX}/{etc,var}/**/*"
    @etc_var_preinstall = Dir[@etc_var_glob]
  end

  def build_bottle_postinstall
    @etc_var_postinstall = Dir[@etc_var_glob]
    (@etc_var_postinstall - @etc_var_preinstall).each do |file|
      Pathname.new(file).cp_path_sub(HOMEBREW_PREFIX, formula.bottle_prefix)
    end
  end
```

So the name shouldn't actually be an issue. Numbering it makes sense, I was just concerned about the mismatch between the formula name (no `@`) and the directory name (with an `@`). But that isn't actually a problem here.

sevan:
Great, lengthy compiles averted all around :)

mistydemeo (approved):
Looking good. I'll merge this batch for now - thank you again for all your work on this!

PR #984, opened 2023-10-26
ruby 2.4.10

sevan:
On the road to switching things to OpenSSL 3, I updated ruby. Unfortunately it doesn't support OpenSSL 3 as is (haven't looked at back-porting patches to `ossl_*.c` files where it fails. 
Otherwise built on Tiger ppc & Leopard x86
It complains about a few things, hence a draft pull request.
```
*** Following extensions failed to configure:
../.././ext/win32/extconf.rb:0: Failed to configure win32. It will not be installed.
../.././ext/win32ole/extconf.rb:0: Failed to configure win32ole. It will not be installed.
```

On Tiger there's a cosmetic issue where it tries to print the summary of configuration at the end of running configure but since it invokes `head` with an unsupported option, it fails to print.

```
Configuration summary for ruby version 2.4.10                                                                                                       
                                                                                                                                                    
head: illegal option -- c                                                                                                                           
Usage: head [-n lines] [file ...]
```

Not sure if this is serious since the build ignores it:
```
making dummy probes.h
/private/tmp/ruby20231026-26902-vefh6v/ruby-2.4.10/tool/vcs.rb:12:in `realpath': undefined method `Pathname' for File:Class (NoMethodError)
        from /private/tmp/ruby20231026-26902-vefh6v/ruby-2.4.10/tool/vcs.rb:108:in `detect'
        from /private/tmp/ruby20231026-26902-vefh6v/ruby-2.4.10/tool/vcs.rb:106:in `loop'
        from /private/tmp/ruby20231026-26902-vefh6v/ruby-2.4.10/tool/vcs.rb:106:in `detect'
        from /private/tmp/ruby20231026-26902-vefh6v/ruby-2.4.10/tool/vcs.rb:104:in `each'
        from /private/tmp/ruby20231026-26902-vefh6v/ruby-2.4.10/tool/vcs.rb:104:in `detect'
        from ./tool/file2lastrev.rb:83
make: [.revision.time] Error 1 (ignored)
```

### DIFF
--- a/Library/Formula/ruby.rb
+++ b/Library/Formula/ruby.rb
@@ -1,9 +1,8 @@
 class Ruby < Formula
   desc "Powerful, clean, object-oriented scripting language"
   homepage "https://www.ruby-lang.org/"
-  url "https://cache.ruby-lang.org/pub/ruby/2.4/ruby-2.4.1.tar.bz2"
-  sha256 "ccfb2d0a61e2a9c374d51e099b0d833b09241ee78fc17e1fe38e3b282160237c"
-  revision 1
+  url "https://cache.ruby-lang.org/pub/ruby/2.4/ruby-2.4.10.tar.bz2"
+  sha256 "6ea3ce7fd0064524ae06dbdcd99741c990901dfc9c66d8139a02f907d30b95a8"
 
   bottle do
   end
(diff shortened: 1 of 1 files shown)


mistydemeo:
The failure to build the win32 extensions should be fine! As for `head`... hm, I wonder if we can just use a newer `head` to build.

Maybe it's worth bumping Ruby past 2.4 at this point so we can do OpenSSL 3.

sevan:
ack, I'll take a look at newer versions of Ruby to see if they're ok with building on Tiger.
I might just patch the configure script to do print the summary by line count rather than add something like coreutils as a build dep, since it's cosmetic.

sevan:
Found the issue causing the crash in `miniruby` and have managed to build the latest release of Ruby 3.2 on Tiger with bodges.
```
% bin/ruby -v
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [powerpc-darwin8.0]
```
Now just need to address the issues I sidestepped :)

sevan:
Of the issues I sidestepped I'm stuck on one issue which needs work. If it's possible this should be merged in the meantime and I will bring in ruby 3 in the new year.

sevan:
Going to merge this version so 2.4.x is up to date until I revisit ruby 3.x since this is done already.

PR #988, opened 2023-11-05
av98: Update test target hostname

sevan:
There's now a dedicated geminiprotocol.net domain.
Use that for testing instead.


mistydemeo (approved):
Nice! Makes sense.

PR #992, opened 2023-11-13
gmp 6.3.0

sevan:
Tested on Tiger (g4/g5) with GCC 4.0.1, Leopard (g4/intel) with GCC 4.2

### DIFF
--- a/Library/Formula/gmp.rb
+++ b/Library/Formula/gmp.rb
@@ -1,18 +1,12 @@
 class Gmp < Formula
   desc "GNU multiple precision arithmetic library"
   homepage "https://gmplib.org/"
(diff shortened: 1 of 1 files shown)


mistydemeo:
Is this compatible with existing things linked against gmp, or will they need to be rebuilt?

sevan:
The [release notes](https://gmplib.org/gmp6.3) state "GMP 6.3 is upwardly source and binary compatible with 6.2, 6.1, 6.0, 5.1, 5.0, 4.x, and 3.x, except for applications that use the semi-documented mpn_bdivmod function.", I only used GCC 5 which seemed fine.

mistydemeo (approved):
Confirmed `brew linkage` seems fine.

PR #996, opened 2023-11-17
re2c 3.1

sevan:
Build tries to use std=c++11 but it really needs gnu++11 with our GCC 5 & 7.

I tried to use `ENV.cxx11` but discovered that it doesn't support GCC 7 (fix is in PR 767), and it doesn't work for this formula since `ENV.cxx11` also sets `-std` to `c++11`.

Tested on Tiger with GCC 5 & 7.


mistydemeo (approved):
Not going to bottle since that'd bake in a specific GCC libstdc++.

sevan:
ok, makes sense.

PR #1010, opened 2023-11-20
libtool 2.4.7

sevan:

Just as a heads up, the builds compiler name is embedded in the `glibtool` script and if that version is not installed, it breaks since `glibtool` tries to invoke that compiler. e.g `gcc-4.2`. Might be worth bottling using 4.0.1 as that will always be installed.


mistydemeo (approved):
> Might be worth bottling using 4.0.1 as that will always be installed.

Good call, done!

PR #1017, opened 2023-11-22
giflib 5.2.1

sevan:
Include patch to Makefile inline since we also need to allow CFLAGS to be overriden, otherwise build breaks with GCC 4.0 since -Wno-format-truncation is not supported.

### DIFF
--- a/Library/Formula/giflib.rb
+++ b/Library/Formula/giflib.rb
@@ -1,41 +1,148 @@
 class Giflib < Formula
   desc "GIF library using patented LZW algorithm"
   homepage "http://giflib.sourceforge.net/"
-  url "https://downloads.sourceforge.net/project/giflib/giflib-4.x/giflib-4.2.3.tar.bz2"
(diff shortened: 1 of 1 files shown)


Library/Formula/giflib.rb
```diff
@@ -1,41 +1,148 @@
...
-    sha256 "3180706f4a94e7ede8c66299474ada34165b2947c262316186e424b1b9d25aba" => :yosemite
-    sha256 "74316a4dd9b94ca052b6f784c9764764d0b24dd8dc1f3f29b5681a374989979a" => :mavericks
-    sha256 "76953a4ac103ff0931f2e4f70dafe9283c9289de2dda7f800e8ca3b47b6830db" => :mountain_lion
   end
 
   option :universal
 
-  depends_on :x11 => :optional
+  # Upstream has stripped out the previous autotools-based build system and their
+  # Makefile doesn't work on macOS. See https://sourceforge.net/p/giflib/bugs/133/
```
mistydemeo, line 14: Well that's annoying!

mistydemeo (approved):

PR #1022, opened 2023-11-24
mercurial 6.6

sevan:
Add gettext as a build dep to get translations
warning: hgbuildmo: could not find msgfmt executable, no translations will be built.

Tested on Tiger (G5) with GCC 4.01


mistydemeo (approved):
Nice we can get rid of the patch!

PR #1025, opened 2023-11-24
whois: Add formula for v5.5.20

sevan:
Based on
https://raw.githubusercontent.com/Homebrew/homebrew-core/f04bca02643e44e5fa8b9fd5aaff131a4d1821b3/Formula/w/whois.rb

Rather that build & copy selected binaries in the original formula, we 'make install' which is where the bash-completion dependency comes in. Opted to add the dependency, rather than patch the Makefile which calls pkg-config to work out where to install bash-completions.

Requires PR #997, #998, #1024 

### DIFF
--- a/Library/Formula/whois.rb
+++ b/Library/Formula/whois.rb
@@ -1,0 +1,45 @@
+class Whois < Formula
+  desc "Lookup tool for domain names and other internet resources"
+  homepage "https://github.com/rfc1036/whois"
+  url "https://ftp.debian.org/debian/pool/main/w/whois/whois_5.5.20.tar.xz"
+  sha256 "42085102dfad82067abe2d5d1cfca59586573dee528718559b022e762bb85cf1"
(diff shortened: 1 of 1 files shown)


Library/Formula/whois.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  bottle do
+  end
+
+  keg_only :provided_by_osx
+
+  depends_on "make" => :build
+  depends_on "perl" => :build
+  depends_on "pkg-config" => :build
+  depends_on "bash-completion"
```
mistydemeo, line 16: Based on your description, is this just a buildtime dependency?

sevan: No, it actually installs bash completion files. If it can't enumerate the path via pkg-config it defaults to `/etc/bash-completion` & fails due to permissions.

mistydemeo (approved):

PR #1036, opened 2023-11-29
openssl: drop makedepend build dep

sevan:
makedepend is slow and the compiler is capable of handling dependencies.
If you have makedepend installed, you should uninstall it on Tiger as we're not guarded against using it and build will take a little longer.

Discovered when investigating why a G4 iBook running Leopard was taking less time than a G5 iMac running Tiger.

Build on iBook G4 with Leopard 39.5 minutes
Build on iMac G5 with Tiger  32.2 minutes


mistydemeo (approved):
Nice catch, thanks!

PR #1039, opened 2023-12-01
utf8proc 2.9.0

sevan:
Built on Tiger (G5) with gcc 4.0.1, Leopard (G4) with gcc 4.2, Snow Leopard, El Capitan with clang.

### DIFF
--- a/Library/Formula/utf8proc.rb
+++ b/Library/Formula/utf8proc.rb
@@ -46,7 +46,7 @@ __END__
  
  libutf8proc.$(MAJOR).dylib: utf8proc.o
 -	$(CC) $(LDFLAGS) -dynamiclib -o $@ $^ -install_name $(libdir)/$@ -Wl,-compatibility_version -Wl,$(MAJOR) -Wl,-current_version -Wl,$(MAJOR).$(MINOR).$(PATCH)
-+	$(CC) $(LDFLAGS) -dynamic -o $@ $^ -install_name $(libdir)/$@ -Wl,-compatibility_version -Wl,$(MAJOR) -Wl,-current_version -Wl,$(MAJOR).$(MINOR).$(PATCH) -dynamiclib
++	$(CC) $(LDFLAGS) -dynamiclib -o $@ $^ -install_name $(libdir)/$@ -compatibility_version $(MAJOR) -current_version $(MAJOR).$(MINOR).$(PATCH)
  
  libutf8proc.dylib: libutf8proc.$(MAJOR).dylib
  	ln -f -s libutf8proc.$(MAJOR).dylib $@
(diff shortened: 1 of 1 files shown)


Library/Formula/utf8proc.rb
```diff
@@ -1,20 +1,16 @@
...
 
   bottle do
-    sha256 "e39f7d87cb5ba31a4f535ac6d01ebfb66e7fe99cc35a99b5398ce584e780c3f9" => :tiger_altivec
   end
 
-  # Unbreak build on legacy compilers which lack warnings for sign conversion
+  # Unbreak build on legacy compilers which lack warnings for sign conversion & linking
   patch :p0, :DATA
 
-  # Uses -compatibility_version, not present in Tiger
```
mistydemeo, line 15: It's still using `-compatibility_version` and passing that straight to `ld`, which is failing for me on Tiger; I think this does still need ld64.

mistydemeo (changes requested):

Library/Formula/utf8proc.rb
sevan: hmm, interesting, I've managed to recreate the issue. I wasn't hitting it because I built it using GCC 4.0, but if I install apple-gcc42, and use that, I hit the issue you mention. Thing is though regardless of either compiler, the same flags are used to build. I looked at the command output and `-Wl,-compatibility_version -Wl,3 -Wl,-current_version -Wl,3.0.0` is passed both times, it just happens to succeed when it's GCC 4.0.
I'll try and dig further to understand why it works on GCC 4.0, my guess is that the error is there but 4.0 just silently ignores it. I'll follow up over the next few days :)

sevan: I typed a reply and pressed resolve conversion, should've pressed comment first.
Found the issue, apple-gcc42 is stricter than 4.0 on Tiger and treats those `-Wl` flags as linker flags where as the build is meant to be using libtool as implied by `-dynamiclib` and `-compatibility_version` is a libtool flag on Tiger. Dropping the `-Wl` flags resolves the issue.
Tested with GCC 4.0 & 4.2 on Tiger ppc, 10.5, 10.6, 10.11 on intel with their default compilers.

mistydemeo: > treats those -Wl flags as linker flags 

Interestingly the manpage for 4.0 still says that `-Wl` passes things to the linker, but I guess there are edge cases that got cleaned up?

mistydemeo (approved):
Thank you for fixing this! Nice to be able to build without ld64 too.

Library/Formula/utf8proc.rb
sevan: Oh, interesting. I have found that 4.0 is definitely much more lenient, there is another scenario in the back of my mind which I can't recall at the moment where there is an issue and 4.0 ploughs through.

sevan: The other thing is that by leopard, `ld` might have supported `-compatibility_version` so GCC 4.2 passing to `ld` would have the desired result.

PR #1042, opened 2023-12-02
gettext 0.22.4

sevan:
This is a biggy since it impacts a lot of packages.
I've built and tested on Tiger with GCC 4.0.1, Leopard GCC 4.2, Snow Leopard, El Capitan clang.


mistydemeo (approved):
Linkage doesn't seem broken in any of the packages I've tested - think we're good without rebuilding!

PR #1043, opened 2023-12-02
poke 3.3

sevan:

Should be merged after #1042 since the new version of gettext includes a library called libtextstyle which poke would like to use and warns about it not being present when built against older versions lacking the library.

### DIFF
--- a/Library/Formula/poke.rb
+++ b/Library/Formula/poke.rb
@@ -1,13 +1,15 @@
 class Poke < Formula
   desc "Extensible editor for structured binary data"
   homepage "https://jemarch.net/poke"
-  url "https://ftp.gnu.org/gnu/poke/poke-3.2.tar.gz"
-  sha256 "758e551dd53a6cce54ec94d8fc21fa4d6b52a27d1c2667206d599ecdc74f0d97"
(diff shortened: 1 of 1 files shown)


mistydemeo (changes requested):
It didn't build for me until I added `libatomic_ops` as a dependency; mind adding that?

mistydemeo:
In my case it tried and failed to link against it until I added the direct dep. Maybe worth adding it to be safe?

sevan:
ah, ok, I've recreated the issue.
The `libatomic_ops` dependency gets added into the poke build process because poke's build uses pkg-config & `Cellar/bdw-gc/7.6.12/lib/pkgconfig/bdw-gc.pc` has `Libs: -L${libdir} -latomic_ops -lgc` but in `bdw-gc.rb` the `libatomic_ops` dependency is marked only for build, so if you've bottled it in a pristine env, the dependency will be missing. I'm just building stuff in place which is why we were getting different results.
With a successful build of poke, libatomic is not linked against in the resulting library nor binaries.

sevan:
Adjusted the libatomic dependency in bdw-gc's formula so hopefully this issue doesn't arise in the future. #1055

mistydemeo (approved):
Tested with GCC 4.2, and still failing, so left that `fails_with`.

PR #1045, opened 2023-12-06
libnghttp2: Add formula for v1.58.0

sevan:
Tested on Tiger (G5) with GCC 4.0.1 as a dependency for curl.

based on
https://github.com/Homebrew/homebrew-core/blob/d60bc504f4526917b7aae5cdae51f188edcc7d8e/Formula/lib/libnghttp2.rb


mistydemeo (approved):
Tested, and confirmed working.

PR #1063, opened 2024-01-30
Add ARM cross compile toolchain

sevan:
arm-none-eabi-binutils: Add a formula for binutils 2.41
Based on
https://github.com/Homebrew/homebrew-core/blob/f5c6266aaf51084a7f9e5d1917cb8f566fee0395/Formula/a/arm-none-eabi-binutils.rb without zstd support for now (needs a C11 compiler + haven't updated formula yet)

Tested on Tiger powerpc (G4/G5) with GCC 4.0.1. Intel: Leopard, Snow Leopard, El Capitan


sevan:
Should've kept it as separate PRs.
Need to sort out a libc for GCC, whether that's newlib or picolib and then they can land.

PR #1064, opened 2024-01-31
libftdi: Need to enable warnings for GCC 4.0

sevan:


mistydemeo (approved):
Confirmed, thank you! I'll hold off on bottling until the libusb stuff lands.

PR #1065, opened 2024-01-31
avrdude 7.0

sevan:
v7.0 is the last autoconf'd version of avrdude, newer versions switch to cmake & require a newer version than we package. Switch to using libftdi 1.x as it is the actively developed release branch.

Tested on Tiger powerpc G4/GCC 7.5 and G5/GCC 5.5, libftid 0.20 & 1.1 and a [littlewire adapter](http://littlewire.github.io).

Requires PR #1064 #1066


sevan:
Working on upgrading deps which should land first.

PR #1066, opened 2024-01-31
libusb 1.0.23

sevan:
This is the last version to support C99. v1.0.24 onwards requires C11 support which requires stock GCC and they struggle with USB.h as there's pragmas defined which are Apple extensions. https://trac.macports.org/ticket/61868#comment:8

Tested on Tiger (G5) with GCC 4.0.1

PR #1070, opened 2024-01-31
flashrom 1.4.0

sevan:
Build tested (I don't think I have any supported hardware at hand) on Tiger (G5) with GCC 5.5

It's likely that flashrom can be compiled with GCC 4.0 but there's a huge struct with an re-occuring issue that GCC 4.0/42 don't like.
Should be merged after https://github.com/mistydemeo/tigerbrew/pull/1066


sevan:
Will need to sort out mason to continue building newer releases that v1.4.0.

PR #1079, opened 2024-02-09
confuse 3.0

sevan:


sevan:
I think if libftdi can build with confuse 3.0, it might be worth setting on that instead, as that's the last version before the use of `fmemopen()` [was introduced](https://github.com/libconfuse/libconfuse/commit/9060af059f5b2e356f820d34caab2cbcc695bf55).

sevan:
I think its best if we stick with v3.0 rather than put up a note as a band-aid for v3.3. libftdi was happy to build against v3.0.

Issue #554, opened 2018-01-24
GCC 7 fails to install on 10.6 Intel at stage3


mistydemeo:
```
warning: gcc/cc1obj-checksum.o differs
warning: gcc/cc1objplus-checksum.o differs
Bootstrap comparison failure!
gcc/cp/semantics.o differs
make[2]: *** [compare] Error 1
make[1]: *** [stage3-bubble] Error 2
make: *** [all] Error 2
```

Logs: https://gist.github.com/a39f7818b6290e7f2a98124a9eb3acfb

GCC 5 and 6 both install fine.

aeggenberger:
I'm getting this one too. Same system and error.

sevan:
Yep, just hit this. Tried to turn on dwarf2 support & it didn't help.

sevan:
I suspect the issue is with one of the configure options in the formula. I tried building the formula without patches, didn't help.
I build GCC manually, outside of tigerbrew, but still used its dependencies from (such as gmp, mpfr) and it managed to build and install.

```
Using built-in specs.
COLLECT_GCC=Sandbox/stock-gcc7/bin/gcc-7
COLLECT_LTO_WRAPPER=/Users/sme/Sandbox/stock-gcc7/libexec/gcc/x86_64-apple-darwin10.8.0/7.5.0/lto-wrapper
Target: x86_64-apple-darwin10.8.0
Configured with: ../configure --prefix=/Users/sme/Sandbox/stock-gcc7 --program-suffix=-7 --with-gmp=/Users/sme/tigerbrew/opt/gmp --with-mpfr=/Users/sme/tigerbrew/opt/mpfr --with-mpc=/Users/sme/tigerbrew/opt/libmpc --with-isl=/Users/sme/tigerbrew/opt/isl --with-system-zlib
Thread model: posix
gcc version 7.5.0 (GCC) 
```

sevan:
> I had a bunch of other experiments in and have cleaned up my tree with just these changes in place for one last build to confirm. Will know in about 3 hours if anything else that I took out was needed. :)

It worked. Just need to turn off compiler optimisation in `gcc.rb` and `Library/Homebrew/extend/ENV/super.rb`

sevan:
I suspect we might be able to work around this by setting `ENV.O0`. It would be nice if the default optimisation settings GCC itself expects could be used on Snow Leopard.

sevan:
[...]

PR #1082, opened 2024-02-16
gcc: Partial build fix for Snow Leopard

sevan:
Optimisation needs to be turned off otherwise build fails on stage 3 comparison. This covers C compiler, todo C++ Compiler.

revision not incremented intentionally to spare lengthy build on working systems.

Tested on 10.6.8 intel with GCC 4.2, 10.8.5 with clang 4.1 (just this change is needed)
Part of #554


sevan:
Fix was cherry picked and landed with the GCC 8 update.

PR #1084, opened 2024-02-20
unbreak `brew test libffi`

gsteemso:
The `test do` block had a bizarrely misconfigured -I include path, which naturally caused the test program to fail with, "hey, what gives, there's no ffi.h file there!".  Fixing the path lets the test program compile, after which the test passes.


mistydemeo:
Thank you! I'd missed this.

Do you mind changing the quotes back to what they were originally? I'd prefer the only change in this PR be to the test.

PR #1085, opened 2024-02-20
Homebrew/extend/fileutils.rb: improve robustness of `make` method

gsteemso:
The `make` convenience method is a great idea, but in the existing configuration cannot handle the case where the "make" package has not been brewed under Tiger (the only place that _needs_ it).

### DIFF
--- a/Library/Formula/libffi.rb
+++ b/Library/Formula/libffi.rb
@@ -1,39 +1,39 @@
 class Libffi < Formula
-  desc "Portable Foreign Function Interface library"
-  homepage "https://sourceware.org/libffi/"
-  url "https://github.com/libffi/libffi/releases/download/v3.4.4/libffi-3.4.4.tar.gz"
-  sha256 "d66c56ad259a82cf2a9dfc408b32bf5da52371500b84745f7fb8b645712df676"
+  desc 'Portable Foreign Function Interface library'
+  homepage 'https://sourceware.org/libffi/'
+  url 'https://github.com/libffi/libffi/releases/download/v3.4.4/libffi-3.4.4.tar.gz'
+  sha256 'd66c56ad259a82cf2a9dfc408b32bf5da52371500b84745f7fb8b645712df676'
 
   head do
-    url "https://github.com/atgreen/libffi.git"
-    depends_on "autoconf" => :build
-    depends_on "automake" => :build
-    depends_on "libtool" => :build
+    url 'https://github.com/atgreen/libffi.git'
+    depends_on 'autoconf' => :build
+    depends_on 'automake' => :build
+    depends_on 'libtool' => :build
   end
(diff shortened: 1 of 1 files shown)


gsteemso:
I only want to merge the one commit here, but the #%*!& Git system won't let me separate them.  Seems like I can massage and mutilate whatever I want locally, but once it's on Github it's permanent?

mistydemeo:
> I only want to merge the one commit here, but the #%*!& Git system won't let me separate them. Seems like I can massage and mutilate whatever I want locally, but once it's on Github it's permanent?

You should be able to remove commits by [rebasing](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History); it's also important to make sure you're using separate branches locally, and not using a single branch to develop from on your computer. That said, I can also cherry-pick just one commit from this PR!

> The make convenience method is a great idea, but in the existing configuration cannot handle the case where the "make" package has not been brewed under Tiger (the only place that needs it).

Good catch. I generally try to ensure that packages that need it have the make dependency specified, but if that gets forgotten it'll be a lot cleaner to actually notify the user like this.

gsteemso:
On Wed, 21 Feb 2024 at 23:19, Misty De Méo ***@***.***> wrote:

> I only want to merge the one commit here, but the #%*!& Git system won't
> let me separate them. Seems like I can massage and mutilate whatever I want
> locally, but once it's on Github it's permanent?
>
> You should be able to remove commits by rebasing
> <https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History>; it's also
> important to make sure you're using separate branches locally, and not
> using a single branch to develop from on your computer. That said, I can
> also cherry-pick just one commit from this PR!
>
Ah, I see!  I shall do so henceforth.

As you can pick out individual files, may I ask you to also pick out the
third one individually?  I had wanted to submit that as a third pull
request during that session, but upon discovering that I could not find any
way to un-smurf-up the parts I had already sent, I was forced to halt the
process early.

> The make convenience method is a great idea, but in the existing
> configuration cannot handle the case where the "make" package has not been
> brewed under Tiger (the only place that needs it).
>
> Good catch. I generally try to ensure that packages that need it have the
> make dependency specified, but if that gets forgotten it'll be a lot
> cleaner to actually notify the user like this.
>
At one point, I was doing development in an environment that had a very
slow edit-recompile loop, so I tend to be hypersensitive to areas that are
guaranteed to fail under easily-deduced conditions.  Thankfully, I haven't
encountered that many so far, but there are still several others in that
particular file and a few others elsewhere.  I shall be submitting other,
similar, pull requests as time permits.

Gordon S.

gsteemso:
Since I plainly failed to keep straight what pull request went with which branch, I am closing and re-submitting them all.

PR #1087, opened 2024-02-22
fix broken `--c++11` option

gsteemso:
(broken regexp common to ENV/std.rb and ENV/super.rb, causing C++11 test to always fail under gcc-7)

### DIFF
--- a/Library/Homebrew/compilers.rb
+++ b/Library/Homebrew/compilers.rb
@@ -2,7 +2,7 @@
 module CompilerConstants
   GNU_GCC_VERSIONS = %w[4.3 4.4 4.5 4.6 4.7 4.8 4.9 5 6 7]
   GNU_GCC_REGEXP = /^gcc-(4\.[3-9]|[5-7])$/
-  GNU_GXX11_REGEXP = /^gcc-(4\.[8-9]|[5-7])$/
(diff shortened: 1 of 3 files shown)


mistydemeo (commented):
I'm not sure if `GNU_GCC_REGEXP` is the right constant here. It covers *all* GCCs in Tigerbrew, from 4.3 onward, but GCC didn't have a full C++11 implementation until 4.8 - which is the version this other regex aims to cover. https://gcc.gnu.org/projects/cxx-status.html

gsteemso:
That makes sense, but it should then be a different declared constant, rather than an explicit value like it is now that will need to be edited in every place it appears whenever you package a new version of GCC.On Mar 24, 2024, at 5:56 PM, Misty De Méo ***@***.***> wrote:﻿
@mistydemeo commented on this pull request.

I'm not sure if GNU_GCC_REGEXP is the right constant here. It covers all GCCs in Tigerbrew, from 4.3 onward, but GCC didn't have a full C++11 implementation until 4.8 - which is the version this other regex aims to cover. https://gcc.gnu.org/projects/cxx-status.html

gsteemso:
I've re-done this so there's a standard, defined-once C++11-aware-GCC regexp right below the standard, defined-once GCC-in-general regexp (in `compilers.rb`), and tweaked stdenv/superenv to use it.

PR #1090, opened 2024-02-27
git 2.47.1

sevan:

git-credential-osxkeychain needs Lion or newer so drop the patch.
regen config.mak.uname patch so it applies again.


sevan:
The `caveats` section should be adjusted so that it only mentions `git-credential-osxkeychain` if it has been built.

PR #1098, opened 2024-03-16
unbreak `brew test libffi`

gsteemso:
For some reason, libffi's formula had a bad path specification in its test method.  This small change fixes it.

### DIFF
--- a/Library/Formula/libffi.rb
+++ b/Library/Formula/libffi.rb
@@ -82,7 +82,7 @@ class Libffi < Formula
      }
     TEST_SCRIPT
 
-    flags = ["-L#{lib}", "-lffi", "-I#{lib}/libffi-#{version}/include"]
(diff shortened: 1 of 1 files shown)


Library/Formula/libffi.rb
```diff
@@ -29,7 +29,7 @@ def install
     system "./autogen.sh" if build.head?
     system "./configure", "--disable-debug", "--disable-dependency-tracking",
                           "--prefix=#{prefix}"
-    system "make", "install"
+    make "install"
```
mistydemeo: This formula doesn't depend on `gmake`, so it shouldn't use the `make` helper. It's only designed for packages that depend on GNU make on Tiger.

mistydemeo (approved):
Thanks! Made one minor change before merging.

PR #1099, opened 2024-03-16
gettext: assorted minor but vital improvements

gsteemso:
make `keg-only` state conditional on OS version; remove unused `--with-examples` option; use the Homebrew `make` routine instead of `system "make"`

### DIFF
--- a/Library/Formula/gettext.rb
+++ b/Library/Formula/gettext.rb
@@ -9,10 +9,11 @@ class Gettext < Formula
     sha256 "c64bb31029e29599442653fe1e0f2216ae8fe144451a0541896f6e367df0018f" => :tiger_altivec
   end
 
-  keg_only :shadowed_by_osx, "OS X provides the BSD gettext library and some software gets confused if both are in the library path."
+  unless MacOS.version <= :leopard  # what Mac OS version would be correct here?
+    keg_only :shadowed_by_osx, "OS X provides the BSD gettext library and some software gets confused if both are in the library path."
+  end
 
   option :universal
-  option 'with-examples', 'Keep example files'
 
   # Fix lang-python-* failures when a traditional French locale
   # https://git.savannah.gnu.org/gitweb/?p=gettext.git;a=patch;h=3c7e67be7d4dab9df362ab19f4f5fa3b9ca0836b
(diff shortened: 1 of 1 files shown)


Library/Formula/gettext.rb
```diff
@@ -9,10 +9,11 @@ class Gettext < Formula
     sha256 "c64bb31029e29599442653fe1e0f2216ae8fe144451a0541896f6e367df0018f" => :tiger_altivec
   end
 
-  keg_only :shadowed_by_osx, "OS X provides the BSD gettext library and some software gets confused if both are in the library path."
+  unless MacOS.version <= :leopard  # what Mac OS version would be correct here?
```
mistydemeo, line 12: Huh, does Tiger *not* include gettext? I don't think I knew that!

Library/Formula/gettext.rb
```diff
@@ -23,32 +24,36 @@ def install
...
-                          "--with-lispdir=#{share}/emacs/site-lisp/gettext",
-                          "--disable-java",
-                          "--disable-csharp",
-                          # Don't use VCS systems to create these archives
-                          "--without-git",
-                          "--without-cvs",
-                          "--without-xz"
-    system "make"
-    system "make", "check"
+    args = %W[
```
mistydemeo, line 27: Mind leaving the args and `make` sections alone in this PR?

gsteemso:
Leopard doesn't either.On Mar 24, 2024, at 5:53 PM, Misty De Méo ***@***.***> wrote:﻿
@mistydemeo commented on this pull request.



In Library/Formula/gettext.rb:
> @@ -9,10 +9,11 @@ class Gettext < Formula
     sha256 "c64bb31029e29599442653fe1e0f2216ae8fe144451a0541896f6e367df0018f" => :tiger_altivec
   end
 
-  keg_only :shadowed_by_osx, "OS X provides the BSD gettext library and some software gets confused if both are in the library path."
+  unless MacOS.version <= :leopard  # what Mac OS version would be correct here?

Huh, does Tiger not include gettext? I don't think I knew that!



In Library/Formula/gettext.rb:
> -                          "--prefix=#{prefix}",
-                          "--with-included-gettext",
-                          "--with-included-glib",
-                          "--with-included-libcroco",
-                          "--with-included-libunistring",
-                          "--with-emacs",
-                          "--with-lispdir=#{share}/emacs/site-lisp/gettext",
-                          "--disable-java",
-                          "--disable-csharp",
-                          # Don't use VCS systems to create these archives
-                          "--without-git",
-                          "--without-cvs",
-                          "--without-xz"
-    system "make"
-    system "make", "check"
+    args = %W[

Mind leaving the args and make sections alone in this PR?

Library/Formula/gettext.rb
gsteemso: I shall do so at my earliest convenience.  Sorry for the mix-up about the purpose of the "make" convenience function.

Library/Formula/gettext.rb
gsteemso: Tiger does not include gettext, nor any fragment thereof.  Neither does Leopard.  I have no idea when it became a standard component, nor indeed whether it ever actually did.

gsteemso: I posted a bug report to that effect months ago.

Library/Formula/gettext.rb
gsteemso: Edits made.  Is it possible to accept the revised PR against HEAD, or do I need to figure out some hocus-pocus to delete the sub-par first attempt that's currently in between them on this branch?

gsteemso:
I wrote an experimental version of the formula for my own use, to upgrade the version and better enable :universal builds, and discovered that the without-tests option HAS to be left out because the gettext package itself no longer offers it.

PR #1100, opened 2024-03-16
extend/fileutils.rb: make the `make` convenience routine more robust

gsteemso:
The `make` convenience method is a great, well, convenience - but the existing implementation breaks under Tiger, which is the only situation where it needs to do anything besides just call the system `make`.  This change fixes that and also improves error reporting.

### DIFF
--- a/Library/Homebrew/extend/fileutils.rb
+++ b/Library/Homebrew/extend/fileutils.rb
@@ -96,17 +96,17 @@ module FileUtils
     system RUBY_BIN/"rake", *args
   end
 
-  # Run `make` 3.81 or newer.
-  # Uses the system make on Leopard and newer, and the
-  # path to the actually-installed make on Tiger or older.
+  # Run `make` 3.81 or newer.  Uses system make from Leopard onward, otherwise brewed make.
   def make(*args)
     if Utils.popen_read("/usr/bin/make", "--version").match(/Make (\d\.\d+)/)[1] > "3.80"
-      system "/usr/bin/make", *args
+      _make = "/usr/bin/make"
+    elsif Formula["make"].installed?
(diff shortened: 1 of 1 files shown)


sevan:
Can you describe the scenario where it breaks on Tiger?

gsteemso:
If brewed `make` is not installed, Tiger’s system `make` is too old to do the job.  The existing implementation not only fails to handle that case, it doesn’t even really check for it very well.

mistydemeo (approved):
Handling an error here seems reasonable, though I might make the `abort` a named exception instead.

The `make` method is only meant to be called in formulae that need it, rather than every package; most packages build with make 3.80, but a few need 3.81 or newer. Those packages have a `make` dependency, but it's still useful to guard against the possibility I forgot to add the dependency even though the `make` method gets used.

gsteemso:
I understand the reasoning there.  I will say that I'm finding whenever I dig into a package that is giving me issues that it tends to expect Make 3.81 or newer, even though they usually don't use any feature that would in fact require it.  Maybe it's just an assumption people now have about modern build environments?  It shouldn't affect anything in the near term, but I suspect that at some point it may become needful to just install the Make package on Tiger the same way you almost _need_ to install Apple-GCC42 if you want anything to build.On Mar 24, 2024, at 5:59 PM, Misty De Méo ***@***.***> wrote:﻿
@mistydemeo approved this pull request.

Handling an error here seems reasonable, though I might make the abort a named exception instead.
The make method is only meant to be called in formulae that need it, rather than every package; most packages build with make 3.80, but a few need 3.81 or newer. Those packages have a make dependency, but it's still useful to guard against the possibility I forgot to add the dependency even though the make method gets used.

PR #1101, opened 2024-03-22
python 3.10.14

sevan:
tcl-tk requires X11 on Tiger and the Python's tkinter module struggles to interact with legacy TCL/TK included with Tiger which has Aqua bindings. So always use the tcl-tk version from tigerbrew.
Use bzip2 from tigerbrew to avoid ancient bundled versions.


mistydemeo (approved):
Thank you!

PR #1104, opened 2024-03-23
id3ed: build fix

sevan:
Assist with finding libhistory and ensure tigerbrew's readline is used.


mistydemeo (approved):
Good catch, thank you!

PR #1106, opened 2024-03-27
:universal option for zlib formula

gsteemso:
trivial edit. appears to work fine with no active intervention.  Have not tried it on Tiger because no way to test the Intel half.

### DIFF
--- a/Library/Formula/zlib.rb
+++ b/Library/Formula/zlib.rb
@@ -9,6 +9,8 @@ class Zlib < Formula
     sha256 "0ec484b96d45d53be8501f85f4b81b2ac2d70d84fd5d1602ece4668d570b05af" => :tiger_altivec
   end
 
+  option :universal
+
   keg_only :provided_by_osx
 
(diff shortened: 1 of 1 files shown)


mistydemeo (commented):
Sorry for the late reply.

Hmm, it doesn't seem to have picked up on it for me - `lipo -info /usr/local/opt/zlib/lib/libz.1.3.1.dylib` still shows a single arch binary. What does that show for you?

gsteemso:
I believe this is due to another issue for which I ought to have submitted a pull request _first_ (I apologize for the confusion).

In brief, Superenv has three methods which modify HOMEBREW_ARCHFLAGS (m32, m64, and universal_binary).  None of them contain the call to `permit_arch_flags` which would make that actually DO anything.

Correcting this deficiency causes the majority (of formulae I have since attempted) to build successfully with this same trivial :universal option.

I suspect that the present PR will work just fine under stdenv, which does not suffer the same deficiency.

PR #1109, opened 2024-04-01
radare2 5.9.0

sevan:
Radare currently splits its Darwin support between architectures so the changes proposed here are for powerpc support.

Based on
https://github.com/Homebrew/homebrew-core/blob/dcbbedf1c0590996b3b76ebc7b9eff92ebc5db73/Formula/r/radare2.rb

Tested on Tiger powerpc (G5) with GCC 5

Should be merged after #1072

### DIFF
--- a/Library/Formula/radare2.rb
+++ b/Library/Formula/radare2.rb
@@ -1,61 +1,135 @@
 class Radare2 < Formula
   desc "Reverse engineering framework"
-  homepage "http://radare.org"
-
-  stable do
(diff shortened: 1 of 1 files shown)


sevan:
Wanted to update this to the latest version (5.9.8) before landing but it looks like the new version uses `CLOCK_MONOTONIC`in `shlr/qjs/src/cutils.c` which Tiger lacks. Sticking with v5.9.0 for now.

Issue #1116, opened 2024-04-04
pcre2 fails to install


noe:
One of the patch files used by the formula has been deleted from the repo it was fetched from.

The patch file URL was https://raw.githubusercontent.com/macports/macports-ports/master/devel/pcre/files/no-OSCacheControl-on-tiger.diff . The file was deleted in commit https://github.com/macports/macports-ports/commit/5c282d8ffea8727f3656213058e52a3d9c688e40 when the pcre2 version was updated to 10.43 (see [commit history](https://github.com/macports/macports-ports/commits/master/devel/pcre/files)).

I understand that the most straightforward fix would be to replace the URL with the URL of the last working version: https://raw.githubusercontent.com/macports/macports-ports/661d0212412d4f428a37d31e233fa6ca1efd4331/devel/pcre/files/no-OSCacheControl-on-tiger.diff . 

Note: pcre2 is a dependency of nmap.

noe:
I created a PR to fix the issue: #1117

sevan:
Thanks for the patch.

PR #1117, opened 2024-04-04
Fix pcre2 non-existing patchfile

noe:
Fixes issue #1116


mistydemeo (approved):
Thank you! Sorry I missed this earlier.

PR #1123, opened 2024-04-05
expat 2.6.2

### LINKED ISSUE #1122: expat 2.5.0 not available
expat's [formula](https://github.com/mistydemeo/tigerbrew/blob/master/Library/Formula/expat.rb) points to version 2.5.0, but that version was withdrawn because it had a vulnerability ([CVE-2023-52425](https://nvd.nist.gov/vuln/detail/CVE-2023-52425)). The tarballs of version 2.5.0 have been removed or renamed from all places, so the download fails during installation.

expat is used by [cmake](https://github.com/mistydemeo/tigerbrew/blob/master/Library/Formula/cmake.rb).

sevan:
Turn off the optional tests as they require a C++11 compiler.

Resolves #1122

### DIFF
--- a/Library/Formula/expat.rb
+++ b/Library/Formula/expat.rb
@@ -1,14 +1,12 @@
 class Expat < Formula
   desc "XML 1.0 parser"
   homepage "http://www.libexpat.org"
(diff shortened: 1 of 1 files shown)


mistydemeo (approved):
Thank you!

PR #1126, opened 2024-04-05
openvpn 2.6.10

sevan:
`enable-password-save` is no longer a recognised `configure` option.
Switch to using openssl3
Build tested on Tiger powerpc (G5) with GCC 4.0.1


mistydemeo (approved):
I'm not able to test, but confirmed it builds fine.

PR #1130, opened 2024-04-05
chibi-scheme 0.10.0

sevan:
Tested on Tiger powerpc (G4 & G5) with GCC 4.0.1


mistydemeo (approved):
Thanks for the work getting this working!

PR #1131, opened 2024-04-06
cc65 2.19

sevan:

### DIFF
--- a/Library/Formula/cc65.rb
+++ b/Library/Formula/cc65.rb
@@ -1,27 +1,17 @@
 class Cc65 < Formula
   desc "6502 C compiler"
-  homepage "https://cc65.github.io/cc65/"
-  # CC65 stable has ceased to be maintained as of March 2013.
-  # The head build has a new home, and new maintainer, but no new stable release yet.
+  homepage "https://cc65.github.io/"
+  url "https://github.com/cc65/cc65/archive/refs/tags/V2.19.tar.gz"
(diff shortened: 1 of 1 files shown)


Library/Formula/cc65.rb
```diff
@@ -1,27 +1,17 @@
...
+  homepage "https://cc65.github.io/"
+  url "https://github.com/cc65/cc65/archive/refs/tags/V2.19.tar.gz"
+  sha256 "157b8051aed7f534e5093471e734e7a95e509c577324099c3c81324ed9d0de77"
   head "https://github.com/cc65/cc65.git"
-  url "ftp://ftp.musoftware.de/pub/uz/cc65/cc65-sources-2.13.3.tar.bz2"
-  sha256 "a98a1b69d3fa15551fe7d53d5bebfc5f9b2aafb9642ee14b735587a421e00468"
 
-  conflicts_with "grc", :because => "both install `grc` binaries"
+  # impressively, Makefile causes ancient make to crash
+  depends_on "make"
```
mistydemeo, line 9: Mind making this `"make" => :build` so it's not a runtime dep?

mistydemeo (changes requested):

Library/Formula/cc65.rb
sevan: done.

mistydemeo (approved):
Thanks!

PR #1134, opened 2024-04-06
python major_minor_version() fix

sevan:
With the recent upgrade from Python 3.7 to 3.10 the minor version needs to be double digits.

Tested by installing Python 3 extensions.
Python 2 untested, marking as draft.

### DIFF
--- a/Library/Homebrew/language/python.rb
+++ b/Library/Homebrew/language/python.rb
@@ -3,7 +3,7 @@ require "utils.rb"
 module Language
   module Python
     def self.major_minor_version(python)
-      version = /\d\.\d/.match `#{python} --version 2>&1`
+      version = /\d\.\d\d/.match `#{python} --version 2>&1`
       return unless version
       Version.new(version.to_s)
     end


sevan:
I think this might be the wrong solution because it breaks Python 2 (end up with directories called python7.18 when 2.7.18 is installed) and needs further investigation.
I found [this commit which introduced support for both python versions](https://github.com/mistydemeo/tigerbrew/commit/0b50110107ea2998e65011ec31ce45931b446dab), just need to see how everything fits in together. I know that I've not had much luck using `:python3` in the past, but haven't worked out why.

sevan:
This is the wrong solution, ended up merging the change to [Library/Homebrew/language/python.rb from upstream](https://github.com/Homebrew/brew/commit/a1efaf1864049977d7d90f32e5ff1e897cf57d96#diff-7e7f927b5a956040b1e5cd4fc54be50d29778145d2cfd57cd1598ed3823e95c0R54) which resolved the issue for installing modules via pip for both python 2.7.18 & 3.10.16.
Will close this and raise a new PR.

PR #1143, opened 2024-04-21
stdenv, superenv: important fixes

gsteemso:
superenv:  added "enable_archflags" to all three places where archflags are
    used (so they actually ARE -- at present everything on my G5 builds for
    ppc7400 no matter what you do), modifying "enable_archflags" to only add
    the K if it isn't already present; and added a pseudo-private conveni-
    ence routine "archflags_enabled?" to aid checking it within super.rb.

both:  added "un_m32" and "un_m64" routines to allow removing them when no
    longer required (important to some :universal workflows), as doing it by
    hand in a way that accounts for both stdenv and superenv is tedious,
    messy, error-prone and fragile;  under superenv this is one of the spots
    where the aforementioned convenience routine comes in handy.

### DIFF
--- a/Library/Homebrew/extend/ENV/std.rb
+++ b/Library/Homebrew/extend/ENV/std.rb
@@ -231,7 +231,7 @@ module Stdenv
     if MacOS::CLT.installed?
       append "CPPFLAGS", "-I/usr/include/libxml2"
     else
-      # Use the includes form the sdk
+      # Use the includes from the sdk
       append "CPPFLAGS", "-I#{MacOS.sdk_path}/usr/include/libxml2"
     end
   end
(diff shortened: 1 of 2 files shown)


gsteemso:
I also altered the routines which account for what CPU is being built on to actually do so, as the only circumstance where that happens now is if you're on intel without ssse4.  The comments say it's left blank because it gets "mutated elsewhere", but neither the ENV/4.3/* build tools nor any file within the Homebrew directory tree do so that I could find.

sevan:
So if I wanted to recreate the issues that you are addressing here what would I need to attempt? (it would be nice to confirm nothing is broken on g3, g4, i386)

gsteemso:
I just realized that I got a small part of this backwards and need to revise it. please hold off on digging into it for a few hours until I am not wasting your time :¬(

gsteemso:
In fact, there are two different things addressed by this PR and it really ought to have been submitted as two PRs.  I'm withdrawing it in order to do so.

PR #1149, opened 2024-04-30
Fix imagesnap for ppc

noe:
Imagesnap's list of valid architectures only includes i386 and x86_64. However, it compiles and works fine in PowerPC. This change adds the ppc and ppc64 architectures in the build invocation.

### DIFF
--- a/Library/Formula/imagesnap.rb
+++ b/Library/Formula/imagesnap.rb
@@ -16,7 +16,7 @@ class Imagesnap < Formula
   depends_on :xcode => :build
 
   def install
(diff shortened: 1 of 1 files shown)


sevan:
Cool!
I wonder if it works with original firewire isight, do you happen to know?

noe:
Sorry, I don't have one of those. I am testing with an old Logitech QuickCam Messenger through [macam](https://sourceforge.net/projects/webcam-osx/).

mistydemeo (approved):
Thank you! I can't test at the moment because I'm on 10.4, but looks good.

PR #1152, opened 2024-05-05
dillo: Add formula for v3.1.0

sevan:
Needs #1151


mistydemeo (approved):
Thanks! I had to revision dump wget for openssl3 at first; having two openssls in the tree meant that dillo was ending up linked to openssl instead of openssl3.

Issue #632, opened 2019-02-24
mplayer on tiger fails to download and install


aIgarcia:
Fresh Tiger installation on PowerBook G4. Xcode 2.5, tigerbrew installed. curl and git installed as well, so far so good.

Python installed fine, however mplayer fails. First it couldn't download the tar.xz due https conflict, I fixed that locally. Then it says /usr/sbin/pkgutil does not exist (and in fact it doesn't). I just copy the pkgutil from a Leopard installation. 

```
libvo/osx_objc_common.m: In function '-[MPCommonOpenGLView preinit]':
libvo/osx_objc_common.m:135: error: 'GLint' undeclared (first use in this function)
libvo/osx_objc_common.m:135: error: (Each undeclared identifier is reported only once
libvo/osx_objc_common.m:135: error: for each function it appears in.)
libvo/osx_objc_common.m:135: error: parse error before 'swapInterval'
libvo/osx_objc_common.m:174: error: 'swapInterval' undeclared (first use in this function)
gmake: *** [Makefile:735: libvo/osx_objc_common.o] Error 1

READ THIS: https://github.com/mistydemeo/tigerbrew/wiki/troubleshooting

```
So I go ahead to the wiki, run brew doctor and get this

`cd /usr/local/Library && git stash && git clean -d -f`

Stashing does nothing, brew doctor keeps showing this but more importantly I'm still unable to install mplayer

kencu:
This might help:
<https://github.com/macports/macports-ports/blob/master/multimedia/MPlayer/files/patch-libvo-osx-objc-common-opengl-headers.diff>

mzryz:
> This might help:
> https://github.com/macports/macports-ports/blob/master/multimedia/MPlayer/files/patch-libvo-osx-objc-common-opengl-headers.diff

Hi kencu, could you please explain how to use this patch in conjunction with the tigerbrew commandline when installing mplayer? Thank you :)

jupo42:
I was able to modify this patch and add it to the mplayer.rb and it successfully passes this part of the build. However, I'm running into further issues with ld errors on the main executable. I'll continue trying to troubleshoot and will post an PR when it's working.

mistydemeo:
Thanks so much! I'd really appreciate a patch. Feel free to open a WIP PR too.

Randrianasulu:
I think patch can be downloaded from
```
[...]

PR #1156, opened 2024-05-13
mplayer 1.5

sevan:
Resolves issue #632
Build is set to use OpenSSL which makes it undistributable as a binary due to license conflict because the gnutls option was unable to stream via HTTPS for me.

Tested on Tiger powerpc (G5) with GCC 4.2, took 16 minutes. Attempted to play mp3 locally, QuickTime movie locally & over network, MP4 h264 over network.

### DIFF
--- a/Library/Formula/mplayer.rb
+++ b/Library/Formula/mplayer.rb
@@ -1,27 +1,28 @@
 class Mplayer < Formula
   desc "UNIX movie player"
-  homepage "https://www.mplayerhq.hu/"
-  revision 1
+  homepage "https://mplayerhq.hu/"
 
   stable do
-    url "http://www.mplayerhq.hu/MPlayer/releases/MPlayer-1.2.tar.xz"
-    sha256 "ffe7f6f10adf2920707e8d6c04f0d3ed34c307efc6cd90ac46593ee8fba2e2b6"
+    url "https://mplayerhq.hu/MPlayer/releases/MPlayer-1.5.tar.xz"
+    sha256 "650cd55bb3cb44c9b39ce36dac488428559799c5f18d16d98edb2b7256cbbf85"
   end
 
   bottle do
-    sha256 "13be908b635ef84ab9a121f7e6787c20e6405bf645417e2f82f98192ea07c92e" => :el_capitan
-    sha256 "c6d4cc95a347807eacf219ed68c6dffae61b5abffae65e938f8e2d959a84c1fb" => :yosemite
-    sha256 "f9c8305916c2eb5363edc7715e499191e18eb39d6f8efd921d3a4d2881326ad9" => :mavericks
   end
 
   head do
     url "svn://svn.mplayerhq.hu/mplayer/trunk"
-    depends_on "subversion" => :build if MacOS.version <= :leopard
-
-    # When building SVN, configure prompts the user to pull FFmpeg from git.
-    # Don't do that.
-    patch :DATA
+    depends_on "subversion" => :build
   end
+
+  # Needs atomic such as sync_bool_compare_and_swap
+  fails_with :gcc_4_0
+
+  # When building SVN, configure prompts the user to pull FFmpeg from git.
+  # help with find OpenGL on Tiger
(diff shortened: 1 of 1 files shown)


Randrianasulu:
yeah, gcc 4.2 was building mplayer-1.5 fine on 10.6.8/x86_64

I thought about how to switch compilers in tigerbrew,  just in case default 4.0.1 is too old but this port probably already does this?

ref: failed functions were apparently some atomic builtins

__sync_val_compare_and_swap

One stackoverflow solution was extract them from new gcc :)
https://stackoverflow.com/questions/9329020/undefined-reference-to-sync-val-compare-and-swap-4-error-at-compilation-usi

Randrianasulu:
I also noticed this recipe started todepend on liba52dec, unfortunately ot seems that upstream tarball at sourceforge is gone?

https://repology.org/project/a52dec/information 

archive.org links exist, not sure if they acceptable here ...

sevan:
> ref: failed functions were apparently some atomic builtins
> 
> __sync_val_compare_and_swap
> 
> One stackoverflow solution was extract them from new gcc :) https://stackoverflow.com/questions/9329020/undefined-reference-to-sync-val-compare-and-swap-4-error-at-compilation-usi

Since mplayer is a leaf package it's safe to just build in with 4.2 rather than perform surgery. :)

sevan:
> I also noticed this recipe started to depend on liba52dec, unfortunately it seems that upstream tarball at sourceforge is gone?

Well spotted, I added it as mplayer can make use of it though I didn't point it to a DVD image in my testing.

> archive.org links exist, not sure if they acceptable here ...

They are, the hash should remain the same and the URL gets updated. Fancy raising a pull request?

Randrianasulu:
@sevan, I prefer not to make myself bigger than I am, right now poor emulated poc still compiling mplayer so I just leave new url for a52 tarball:

https://web.archive.org/web/20170110045630if_/http://liba52.sourceforge.net/files/a52dec-0.7.4.tar.gz

sadly it is https

Randrianasulu:
I also want to say that nontigerbrew build of MPlayer 1.5 finished some time ago, just for some reason imf and dash demuxers inside ffmpeg were non-compileable with libxml2 , so I disabled them.

sevan:
Do you mean that you built with the same options & dependencies outside of Tigerbrew & it was quicker or that with all the added dependencies in this formula, it is taking a lot longer to build here?

Randrianasulu:
> Do you mean that you built with the same options & dependencies outside of Tigerbrew & it was quicker or that with all the added dependencies in this formula, it is taking a lot longer to build here?

====

I pre-installed some deps I saw in recipe, some of them were binary already pre-compiled ("bottles"). But my qemu setup mostly run as 32-bit binary on host, and single-threaded on top of that (some unemulated piece of mac hardware - I am looking at it with qemu developers). So, for me everything is much slower than on real machine.

I already had mplayer sources unpacked, so it was just matter of running configure/make until binary finally appear. I wonder why it builds for you  but not for me with tigerbrew-provided libxml2?
Anyway, I think I have brew update issue solved thanks to this issue tracker, so I'll retry 'correct way' as fast as this MR hit main repo :)

sevan:
> just for some reason imf and dash demuxers inside ffmpeg were non-compileable with libxml2 , so I disabled them.

Without any error message I don't have much to go on since I never hit the issue. I've raised a pull request to update libxml2 from 2.11.4 to 2.11.8 if you want to give that a try. What compiler did you use by the way? I used the apple-gcc42 compiler.

Randrianasulu:
@sevan I also used apple-gcc42 (by setting cc variable manually). May be  initially build failed for me   because I failed to consider setting pkg_config variable this manual way too?

I thought about using distcc next time, between two virtual machines running on different hosts. In theory Xcode installation should contain distcc already, but I read somewhere it might be hardcoded for Apple's gcc? Any expirience in this arrea?

sevan:
> @sevan I also used apple-gcc42 (by setting cc variable manually). May be initially build failed for me because I failed to consider setting pkg_config variable this manual way too?

depends on the error, but it would help.

> I thought about using distcc next time, between two virtual machines running on different hosts. In theory Xcode installation should contain distcc already, but I read somewhere it might be hardcoded for Apple's gcc? Any expirience in this arrea?

I've not, sounds like fun :)

PR #1161, opened 2024-05-15
lftp 4.9.2

### LINKED ISSUE #1157: lftp fails to build on 10.4 ppc?
Installed tigerbrew on second virtual machine, installed some libs like libmpeg2 :)

Tried to install lftp and it errored out in ssl support, even if includes were from /usr/local/opt, i.e. not outdated?

Screenshot_20240515-210714

sevan:
Need bison because it mark older version found on Tiger as bad Switch to OpenSSL 3
Add dependencies which it can make use of
Fix test
Needs a C++11 compiler since it makes use of override specifier
Tested on Tiger powerpc (G5) with GCC 5.5

Resolves #1157 

Marking as draft until I find out if it affects more recent versions of OS X & how to deal with the issue (__WORDSIZE definitely isn't used in the codebase apart from in these 2 files)

### DIFF
--- a/Library/Formula/lftp.rb
+++ b/Library/Formula/lftp.rb
@@ -1,28 +1,68 @@
 class Lftp < Formula
   desc "Sophisticated file transfer program"
   homepage "http://lftp.yar.ru/"
-  url "http://lftp.yar.ru/ftp/lftp-4.6.4.tar.xz"
-  sha256 "1e7cb674c83ace48172263f86847ed04bb6ab2f24116b11a8505f70a15e8805c"
+  url "http://lftp.yar.ru/ftp/lftp-4.9.2.tar.xz"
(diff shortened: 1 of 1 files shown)


mistydemeo:
libidn2 is ready now. I don't have access to later non-Homebrew-supported Macs to test those other things though.

sevan:
No worries, I'll try and get to it before the weekend.

sevan:
fought with this on Snow Leopard+GCC 7 and haven't had much luck, it has its own autoconf module for detecting readline and it struggles with setting CFLAGS correctly (setting it to the equivalent of `readline.opt_include/readline` but in source including from `<readline/readline.h>`. Now wondering how it worked on Tiger with GCC 5 (where Tigerbrew is installed is different on the two systems).

Also had to change Tigerbrew to recognise GCC 7 as a C++11 compiler but that's a separate issue. :)

Will come back to this again, later.

PR #1164, opened 2024-06-01
fpc 3.2.2 / 3.0.4

sevan:
Freepascal 3.2.2 needs a minimum version of 3.0.4 to bootstrap. v3.2.2 needs 10.9 or newer to build on Intel darwin. It tries to target 10.8 but fails.
It is not possible to build the Intel target compiler on PowerPC. So build the latest version on PowerPC systems & those running >=10.9. For 10.4 to 10.8 Intel builds stick with 3.0.4.

Tested on Tiger powerpc (G5) with GCC 4.0.1, Snow Leopard with clang.

Draft: needs more testing, test target in formula needs to be fixed as it currently hardcodes `powerpc-darwin` in path.

### DIFF
--- a/Library/Formula/fpc.rb
+++ b/Library/Formula/fpc.rb
@@ -1,19 +1,22 @@
 class Fpc < Formula
   desc "Free Pascal: multi-architecture Pascal compiler"
   homepage "http://www.freepascal.org/"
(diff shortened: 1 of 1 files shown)


sevan:
Fails on OS X versions which support code signing (Lion and newer?) since it tries to sign the produced binary.
Hit it on El Capitan.

mistydemeo:
Hmm, that codesigning bug is a fun one...

PR #1077, opened 2024-02-06
xz 5.4.7

sevan:
Tested on Tiger powerpc (G5) with GCC 4.0.1

PR #1169, opened 2024-06-07
libarchive 3.7.4

sevan:
Use up to date libraries from Tigerbrew as dependencies
Build only checks for the presence of `sys/queue.h` and not whether the actual required macros are defined. 
Use the up to date copy of queue.h included in libarchive instead.

Should be merged after #1077


mistydemeo (approved):
Thank you!

PR #1171, opened 2024-06-10
perl 5.40.0

sevan:
Tested on Tiger powerpc (G5) with GCC 4.0.1., Leopard powerpc (G4) with GCC 4.2

Took 108.3 minutes on a 1.8Ghz iMac G5. 132mins on a 1.33Ghz iBook G4.


mistydemeo (approved):
110.9 minutes on a G5 PowerMac. Seems like the tests are the vast majority of the time.

PR #1176, opened 2024-06-20
zzuf 0.15

sevan:
There is no longer a conflict as a zzat command is installed now
Fix test (copied from upstream Homebrew formula)

Tested on Tiger powerpc (G5) with Apple-GCC 4.2.


mistydemeo (approved):
Thank you!

PR #1180, opened 2024-07-02
openssh 9.8p1

mistydemeo:


mistydemeo:
Patches don't currently apply.

mistydemeo:
Updated patch - also upstreamed it to Homebrew.

https://github.com/Homebrew/homebrew-core/pull/176149 / https://github.com/Homebrew/formula-patches/pull/922.

PR #1191, opened 2024-09-25
Macvim v7.4.909

rdustinb:
Updating MacVim to the latest possible tagged version (tag 80) that can build on OSX 10.5.8. No version later than this seem to build correctly.

### DIFF
--- a/Library/Formula/macvim.rb
+++ b/Library/Formula/macvim.rb
@@ -104,14 +104,14 @@ class Macvim < Formula
 
     system "./configure", *args
 
-    #if build.with? "custom-icons"
-    #  # Get the custom font used by the icons
-    #  system "make", "-C", "src/MacVim/icons", "getenvy"
-    #else
-    #  # Building custom icons fails for many users, so off by default.
-    #  inreplace "src/MacVim/icons/Makefile", "$(MAKE) -C makeicns", ""
-    #  inreplace "src/MacVim/icons/make_icons.py", "dont_create = False", "dont_create = True"
-    #end
+    if build.with? "custom-icons"
+      # Get the custom font used by the icons
+      system "make", "-C", "src/MacVim/icons", "getenvy"
+    else
+      # Building custom icons fails for many users, so off by default.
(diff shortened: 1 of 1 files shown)


sevan:
The pending conflict which needs to be resolved aside, I was able to get it to compile on Tiger with GCC 4.0 after I skipped on Ruby, Python2 & 3 support.
Didn't investigate the failures but I suspect the autoconf parts need to be looked as it is assuming older version where as I was building against Python 2.7.10, 3.10, Ruby 3.2.2.

Looks like we need to source a fresh set of icons too (see dock in screenshot).

Picture 14

sevan:
With the [`icns` files from MacVim snapshot-70 branch ](https://github.com/macvim-dev/macvim/tree/snapshot-70/src/MacVim/icons)copied into place, I get a working dock icon.

Picture 15

rdustinb:
@sevan thank you for testing the build. Since it has been a while, it seems like I should sync my fork and branch, then resubmit the pull request? Or does any pull request that is open update with any changes to the branch underneath?

~~Also it seems like I need to copy the icns into my branch to include them as part of the pull request?~~ I'm an idiot, I forgot the source isn't in the tiger brew repo, sounds like the macvim formula needs to use "custom icons" and point to snapshot-70, or maybe there is a better way to handle this automatically?

Thank you for your help!

sevan:
> @sevan thank you for testing the build. Since it has been a while, it seems like I should sync my fork and branch, then resubmit the pull request? Or does any pull request that is open update with any changes to the branch underneath?

There's a conflict which needs to be resolved, if you rebase and resolve the conflict that will allow your current changs to land.

> ~Also it seems like I need to copy the icns into my branch to include them as part of the pull request?~ I'm an idiot, I forgot the source isn't in the tiger brew repo, sounds like the macvim formula needs to use "custom icons" and point to snapshot-70, or maybe there is a better way to handle this automatically?

I guess you could use the URL for the icon and define a new resource and in the install block of the formula, overwrite the existing icon in the extracted source? (In theory, I haven't tried it)
Let me know if the experiment fails, and I'll dig in further.

PR #1037, opened 2023-11-29
openssl 3.4.0

sevan:
It's now possible to build on Tiger with asm support. `OPENSSL_LOCAL_CONFIG_DIR` landed in OpenSSL 1.1.0 and we've building with it untouched in `openssl.rb`, don't bother unsetting it here too.
Skip setting compiler optimisation, the build ends up setting -O3 in places.
By making the test suite run optional, and only installing manuals, build time is significantly reduced.

120minutes before (full docs, suite run) to
no test, with docs, 35.9 minutes
no test, with man, 29.3
Building no tests, no docs results in 19.5mins on 1.8Ghz G5 iMac.

224 minutes with tests & man, 44minutes no tests, with  man, on a 1.33Ghz G4 PowerBook.
50.5 minutes no tests, with man, on a 1.33Ghz iBook G4

Tested on Tiger powerpc (G4/G5) and intel (core2duo) with GCC 4.0.1, Leopard powerpc (G4) and intel (core2duo) with GCC 4.2, El Capitan with llvm

PR #1193, opened 2024-10-14
openssh 9.9p1

sevan:
Drop the sandbox patches since with the further process separation changes which landed in v9.8, sshd breaks.
macOS 10+5 ships with v9.8 and the source drop also shows they are no longer applying these patches.
See OpenSSH-341 sources:
https://github.com/apple-oss-distributions/OpenSSH/tree/OpenSSH-341/

Tested on Tiger powerpc (G5) with GCC 4.0.1, Leopard powerpc (G4) with GCC 4.2

### DIFF
--- a/Library/Formula/openssh.rb
+++ b/Library/Formula/openssh.rb
@@ -1,14 +1,13 @@
 class Openssh < Formula
   desc "OpenBSD freely-licensed SSH connectivity tools"
   homepage "https://www.openssh.com/"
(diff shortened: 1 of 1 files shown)


mistydemeo:
Nice find, thank you! I'll give this a test soon.

sevan:
Ideally should be merged after #1037 to bring in an up to date version of openssl, first.

PR #1210, opened 2024-12-26
libassuan 3.0.1

sevan:


mistydemeo (approved):
Ended up needing #1217 to make sure libassuan installs `assuan-config`.

PR #1218, opened 2024-12-28
recode 3.7.14

sevan:
Switch to a fork which is actively maintained.
The original repo hasn't been touched in some years and has build issues.
This fork is also used by MacPorts.

Tested on Tiger (G5) with GCC 5.


mistydemeo (commented):
Hmm, not building with GCC 4.0 and 4.2. Should this be marked as incompatible with them?

mistydemeo:
Wait, no - it worked with 4.0, just not 4.2.

mistydemeo (approved):

PR #1219, opened 2024-12-29
ld64: Use OpenSSL included with OS

sevan:
Saves a detour into building OpenSSL + deps when version included with OS is sufficient for providing MD5 support.

Tested on Tiger (G5) with GCC 4.2, then built `cctools` which depends on `ld64`

Not sure if a revbump is needed as it will create churn, really beneficial for those on G3 based systems, starting out, who do not get prebuilt packages in the first place.

```
Cellar/ld64/97.17_1/bin/ld:
	/usr/lib/libcrypto.0.9.7.dylib (compatibility version 0.9.7, current version 0.9.7)
	/usr/lib/libstdc++.6.dylib (compatibility version 7.0.0, current version 7.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 88.1.12)
```

### DIFF
--- a/Library/Formula/ld64.rb
+++ b/Library/Formula/ld64.rb
@@ -26,8 +26,6 @@ class Ld64 < Formula
   depends_on "cctools-headers" => :build
   depends_on "dyld-headers" => :build
   depends_on "libunwind-headers" => :build
-  # No CommonCrypto
-  depends_on "openssl" if MacOS.version < :leopard
 
   keg_only :provided_by_osx,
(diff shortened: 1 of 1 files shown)


sevan:
> Not sure if a revbump is needed as it will create churn, really beneficial for those on G3 based systems, starting out, who do not get prebuilt packages in the first place.

hmm, on the other hand, one less package which installs the openssl package...
I leave it to you to decide. :)

mistydemeo (approved):
I'll prep a new bottle with this.

PR #1225, opened 2025-01-05
python 3.10.16

sevan:
Skipping upgrading pip as setup.py was removed in the following release. Expectation is ensurepip module is used now with the newer releases. Add gettext as dependency since version included in 10.5 is missing symbols

Undefined symbols:
  "_libintl_textdomain", referenced from:
      __locale_textdomain in libpython3.10.a(_localemodule.o)
  "_libintl_dgettext", referenced from:
      __locale_dgettext in libpython3.10.a(_localemodule.o)
  "_libintl_dcgettext", referenced from:
      __locale_dcgettext in libpython3.10.a(_localemodule.o)
  "_libintl_bindtextdomain", referenced from:
      __locale_bindtextdomain in libpython3.10.a(_localemodule.o)
  "_libintl_gettext", referenced from:
      __locale_gettext in libpython3.10.a(_localemodule.o)

Tested on Tiger (G5/c2d) with GCC 4.0.1, Leopard (G4) with GCC 4.2

### DIFF
--- a/Library/Formula/python3.rb
+++ b/Library/Formula/python3.rb
@@ -1,11 +1,10 @@
 class Python3 < Formula
   desc "Interpreted, interactive, object-oriented programming language"
   homepage "https://www.python.org/"
-  url "https://www.python.org/ftp/python/3.10.14/Python-3.10.14.tar.xz"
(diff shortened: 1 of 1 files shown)


sevan:
This should be bottled after the libffi update in #1227

PR #1237, opened 2025-01-13
python3: Switch to using libffi formula

sevan:
With the upgrade to 3.10.16, builds on a Tiger G4 resulted in the ctype module not building. It turns out that it libffi was being linked against opportunistically via pkg-config in some scenarios. Just depends on libffi since the version found in the source of Python (Modules/_ctypes/libffi_osx) is very old (a pre-release of 2.0).

Some modules also link against zlib so depend on that too. Add a test to ensure the ctype module works in that it imports as the build of module can fail and the build of Python can succeed only to find that there no module to import later.
While here add a guard for tkinter test since only on Xorg X11 in 10.5 and up will X11 be started and the test run within. On Tiger you have in invoke the test from xterm otherwise it'll just fail.

Tested on Tiger (G4/G5/i386) with GCC 4.0.1

Build on a 1.3Ghz G4 PowerBook 19 minutes, 1.8Ghz iMac G5 15 minutes, 2Ghz c2d Mac Mini 4 minutes.

### DIFF
--- a/Library/Formula/python3.rb
+++ b/Library/Formula/python3.rb
@@ -3,6 +3,7 @@ class Python3 < Formula
   homepage "https://www.python.org/"
   url "https://www.python.org/ftp/python/3.10.16/Python-3.10.16.tar.xz"
   sha256 "bfb249609990220491a1b92850a07135ed0831e41738cf681d63cf01b2a8fbd1"
+  revision 1
 
   bottle do
   end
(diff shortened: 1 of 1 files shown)


sevan:
Forgot to mention about what to check.
`otool -L Cellar/python3/3.10.16_1/Frameworks/Python.framework/Versions/3.10/lib/python3.10/lib-dynload/_ctypes.cpython-310-darwin.so`

For zlib, `binascii.cpython-310-darwin.so` & the zlib module itself (`zlib.cpython-310-darwin.so`)

PR #1240, opened 2025-01-14
flac 1.4.3

### LINKED ISSUE #335: FLAC failed to build on 10.5 Intel
There are several linker errors when trying to build FLAC with `brew install --with-libogg flac`
They are caused by `nasm` failure to compile assembly files. Installation on the latest version of `nasm` didn't help.
Also, passing `--enable-sse` in the formula seems kinda odd. I didn't try to build for PowerPC though.

sevan:
Drop universal build support as build was broken on new version.
`--enable-sse` no longer recognised.
configure sets things up, no need to set [OBJ_FORMAT](url), assembly support is enabled if building with clang.
Due to shortcoming in clang shipped with 10.8/10.9, llvm-gcc needs to be used, but there's no longer an issue building with llvm-gcc.
Enable OGG support, build times are for both are very short.

Tested on Tiger (G5/G4/i386) with GCC 4.0.1, Leopard (G4/i386) with GCC 4.2, 10.6 (i386) with GCC 4.2 / llvm-gcc, 10.7 & 10.8 with clang/llvm-gcc, 10.9 to 10.11 with clang.
Resolves #335

### DIFF
--- a/Library/Formula/flac.rb
+++ b/Library/Formula/flac.rb
@@ -1,8 +1,18 @@
 class Flac < Formula
   desc "Free lossless audio codec"
   homepage "https://xiph.org/flac/"
-  url "http://downloads.xiph.org/releases/flac/flac-1.3.1.tar.xz"
-  sha256 "4773c0099dba767d963fd92143263be338c48702172e8754b9bc5103efe1c56c"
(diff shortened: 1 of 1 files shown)


sevan:
Made a mistake in the commit message, the shortcoming was in the version of clang shipped with 10.7 & 10.8, not "10.8/10.9". Builds find with clang on 10.9.

PR #1246, opened 2025-01-15
pilot-link: Add formula for v0.12.5

sevan:
Marking as daft because it needs more testing.
Works fine on Tiger because there's no native drivers for USB sync, but on Leopard there is and trying to access the device via `libusb` fails. Need to make that conditional for when on Tiger & test some more, then it can land.

### DIFF
--- a/Library/Formula/pilot-link.rb
+++ b/Library/Formula/pilot-link.rb
@@ -7,7 +7,6 @@ class PilotLink < Formula
   sha256 "d3f99ec04016b38995fb370265200254710318105c792c017d3aaccfb97a84b2"
 
   depends_on "pkg-config" => :build
-  depends_on "libusb-compat"
   depends_on "libpng"
   depends_on "popt"
   depends_on "readline"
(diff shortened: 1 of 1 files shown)


sevan:
Leopard comes with drivers for the USB docking station which causes
a conflict when attempting to use pilot-link built with libusb support.
Turns out pilot-link can communicate with PDAs via IOKit just fine on
Tiger and other OS X versions without drivers for Palm so just use
that.
Tested on Tiger with a Palm m515 & USB dock + Palm Vx, RS-232 dock,
Prolific PL2303 USB to RS-232 adapter.
Leopard to El Capitan just via USB dock.

Tested on Tiger PowerPC (G5) with GCC 4.0.1, Leopard (G4/i386) with GCC 4.2, Snow Leopard to El Capitan with respective default compilers.

sevan:
Missed a key word in the commit message.
Meant to say
"Turns out pilot-link can communicate with PDAs via IOKit just **fine** on
Tiger and other OS X versions without drivers for Palm so just use
that."
but instead said
"Turns out pilot-link can communicate with PDAs via IOKit just on
Tiger and other OS X versions without drivers for Palm so just use
that."

PR #1247, opened 2025-01-16
qt: fails with gcc-4.2 on Tiger

mistydemeo:
@sevan Not sure if you tested with GCC 4.2 on Tiger, or just 4.0 - wanted to double check if this lines up with what you'd seen.


sevan:
Hi @mistydemeo 
Doesn't look like I tried with GCC 4.2. Stuck with 4.0.1.

PR #1256, opened 2025-01-19
gmp: Drop workaround for OS X 10.6

sevan:
Workaround is on longer required on 10.6 and slows down builds for all OS X versions.
Bump revision as assembly support is no longer explicitly disabled.

Add a test for static libraries from upstream
https://github.com/Homebrew/homebrew-core/blob/65f68fef5c44448232c8dab8b748baf33e8daf61/Formula/g/gmp.rb

Tested on OS X 10.6 as a 64bit build & 32bit build, with GCC 4.2 & llvm-gcc.


mistydemeo (approved):
Makes sense! Thanks for adding the test. I'm away from the G5 today, but I can bottle the new revision tomorrow.

sevan:
No rush, just wasn't sure if it would mess with the plans for incoming gcc upgrades.

PR #1260, opened 2025-01-29
sqlite: Use llvm-gcc on 10.7 & 10.8

sevan:
`___atomic_load_n()` / `___atomic_store_n()` called but not defined.

### DIFF
--- a/Library/Formula/sqlite.rb
+++ b/Library/Formula/sqlite.rb
@@ -39,6 +39,9 @@ class Sqlite < Formula
     sha256 "62e51962552fb204ef0a541d51f8f721499d1a3fffae6e86558d251c96084fcf"
   end
 
+  #___atomic_load_n() / ___atomic_store_n() called but not defined.
+  fails_with :clang if MacOS.version == :lion || MacOS.version == :mountain_lion
+
   def install
     # sqlite segfaults on Tiger/PPC with our gcc-4.2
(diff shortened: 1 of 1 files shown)


mistydemeo (approved):

mistydemeo (commented):
Curious, what clang build range is it failing in?

sevan:
Looks like it's 421.11.66 on both OS releases.

Lion:
```
Apple clang version 4.1 (tags/Apple/clang-421.11.66) (based on LLVM 3.1svn)
Target: x86_64-apple-darwin11.4.2
Thread model: posix
```

Mountain Lion:
```
Apple clang version 4.1 (tags/Apple/clang-421.11.66) (based on LLVM 3.1svn)
Target: x86_64-apple-darwin12.6.0
Thread model: posix
```

mistydemeo:
In that case, maybe we can update the `fails_with` to be block-form? That lets you specify the cause as part of the `fails_with`, for example: https://github.com/mistydemeo/tigerbrew/blob/8ba134eda537d2cee7daa7ebdd9f728389d9c53e/Library/Formula/ape.rb#L14-L17

It's not a big deal, just figure it's nice to make use of the metadata when we can.

sevan:
Makes sense, it'd also allow a newer build of clang to be used in the future if we have that going.
I'll make the change tomorrow and merge it.

PR #1288, opened 2025-02-18
gnupg: doesn't build with certain older compilers

mistydemeo:
cc @sevan 

I'm not sure if we also need to pick a clang build, too; I'm not currently using any systems that have a clang old enough that it wouldn't be able to build this.

### DIFF
--- a/Library/Formula/gnupg.rb
+++ b/Library/Formula/gnupg.rb
@@ -13,6 +13,11 @@ class Gnupg < Formula
 
   depends_on "curl" if MacOS.version <= :mavericks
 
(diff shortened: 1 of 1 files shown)


sevan:
I'll try it out tomorrow & report back.

mistydemeo:
OK, I'll leave it at these for now then!

sevan:
Sorry, I wasn't paying close enough attention when reviewing. We have an off-by-two error! ;) (the exclusions need to be in `gnupg2.rb` not `gnupg.rb`)

mistydemeo:
Oops!! You're totally right, I'll revert this.

PR #1298, opened 2025-02-26
gettext/libpng/sqlite: revert clang exclusion

sevan:
The issues with clang on Lion and Mountain Lion were due to my own mistake of having outdated command line tools installed.
With the latest version installed, there's no issue on Lion with clang-425.0.28 and Mountain Lion with clang-503.0.40.


mistydemeo:
Good point! I didn't consider that this might be specific to the older CLT you were using.

sevan:
Search results on http://developer.apple.com/download/all/ are pretty mixed up.
I assumed top most hit was the most recent. :(

PR #1300, opened 2025-02-27
GCC 4.9.4

sevan:
Add a formula for isl 0.12.2
Unlike GCC 4.8 which requires isl 0.11.1 as a minimum, GCC 4.9 states that isl 0.12.2 is required in its install documentation.

GCC 4.9.4
Drop patches, they were specifically for 4.9.3. 4.9.4 was listed as
unaffected.
Need isl 0.12.2 to build GCC 4.9.4
Turn off optimisation to avoid compilation issues during a lengthy build.
--enable-libstdcxx-time alters the C++11 ABI and shouldn't be used. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61758#c6
Allow the build itself to set the type of checking.
Use the default build config & address any fallouts, rather than use a debug bootstrap.
We're not building head so no need to specify disable werror.
Update URLs

Tested on Tiger (G4) with GCC 4.0.1
Built on a 1.33Ghz PowerBook G4 running Tiger in 481 minutes using GCC 4.0.1
Built on a 1.8Ghz iMac G5 running Tiger in 438 minutes using GCC 4.0.1
Built on a 2Ghz c2d mac mini running Tiger in 84 minutes using GCC 4.0.1
Built on a 1.33Ghz iBook G4 running Leopard in 327 minutes using GCC 4.2
Built on a 1.8Ghz iMac G5 running Leopard in 247 minutes using GCC 4.2
Built on a mid-2009 Air running Leopard using GCC 4.2 in 89 minutes.
Built on a mid-2009 Air running Snow Leopard using GCC 4.2 in 113 minutes.
Built on late 2009 white MacBook running Lion using clang in 60something minutes.
Built on late 2009 white MacBook running Mountain Lion using clang in 67 minutes.
Built on late 2009 white MacBook running Mavericks using clang in 73 minutes.
Built on late 2009 white MacBook running Yosemite using clang in 104 minutes `--with-build-config=bootstrap-debug` (only way to get it to build)
Built on a mid-2009 Air running El Capitan using clang in 112 minutes.
Marking as draft until I've done more testing beyond just Tiger

### DIFF
--- a/Library/Formula/gcc49.rb
+++ b/Library/Formula/gcc49.rb
@@ -21,32 +21,13 @@ class Gcc49 < Formula
 
   desc "The GNU Compiler Collection"
   homepage "https://gcc.gnu.org"
-  url "https://ftpmirror.gnu.org/gcc/gcc-4.9.3/gcc-4.9.3.tar.bz2"
-  mirror "https://ftp.gnu.org/gnu/gcc/gcc-4.9.3/gcc-4.9.3.tar.bz2"
-  sha256 "2332b2a5a321b57508b9031354a8503af6fdfb868b8c1748d33028d100a8b67e"
+  url "https://ftpmirror.gnu.org/gcc/gcc-4.9.4/gcc-4.9.4.tar.bz2"
+  mirror "https://ftp.gnu.org/gnu/gcc/gcc-4.9.4/gcc-4.9.4.tar.bz2"
+  sha256 "6c11d292cd01b294f9f84c9a59c230d80e9e4a47e5c6355f046bb36d4f358092"
 
   head "svn://gcc.gnu.org/svn/gcc/branches/gcc-4_9-branch"
 
   bottle do
-    rebuild 3
-    sha256 "0acf2b010e3c2210fcb5fdc03de9dc14f0556cf5295347d8f72f2d8d1cfab4ff" => :el_capitan
-    sha256 "5b635a24e9f7464fb94d2933b00a8d1b1cfc0efb1de140fe568e313d24965140" => :yosemite
-    sha256 "dc24f86a9652fbb0ec0bc9dd0103d23bb68c315bd490ee5d0b10a7144453ecc4" => :mavericks
-  end
-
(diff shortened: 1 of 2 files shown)


Library/Formula/gcc49.rb
```diff
@@ -80,6 +61,7 @@ def pour_bottle?
   cxxstdlib_check :skip
 
   def install
+    ENV.no_optimization
```
mistydemeo, line 64: Did this break in all circumstances, or did it depend on what the bootstrapping compiler was?

sevan: This is the first time I've gone back to trying to compiler compilers on Tigerbrew with GCC 4.0.1 on Tiger.
Previous attempts have always started with 4.2.
I was able to build 4.4 & 4.5 on Leopard/PowerPC without issue using GCC 4.2 however the issue you ran into with the stage comparison is masked by using `--with-build-config=bootstrap-debug` where as it is related to optimisation.
https://github.com/Homebrew/legacy-homebrew/pull/38596
I'll be able to give a proper explanation and fix once I've done more testing.
I switched it off as 4.8 & 4.9 builds blow up late in the process and waste a very large chunk of time +  I wanted to spare my PowerBook. :)

sevan: I don't have a fix since I can't really ruby and what I tried didn't work, but it turns out that we always build with `-Os` via the superenv & the `cc` ENV wrappers. All setting `no_optimization` really does is ensures that `-march` is not applied. (check the log file end with `.cc` for a package to see superenv performing the substitution.
It would be

sevan: It seems all formulas from GCC 4.5 to 14 need optimisation turned off. I just did it pre-emptively on this and 4.8's formula.

sevan: ok to land it?

mistydemeo: That's fine, go for it!

PR #1305, opened 2025-03-11
gcc47: tweak build

sevan:
--enable-libstdcxx-time alters the C++11 ABI and shouldn't be used. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61758#c6
Allow the build itself to set the type of checking.
We're not building head so no need to specify disable werror.
Update URLs
Don't blindly enable LTO, it is not supported on PowerPC https://github.com/iains/darwin-toolchains-start-here/discussions/37#discussioncomment-3285847
Add a test for C++ & Fortran
Skip the conditional as the fix from the GCC bug was integrated into GCC.

Built on a 1.33Ghz PowerBook G4 running Tiger using GCC 4.0.1 in 333 minutes
Built on 1st gen 1.8Ghz iMac G5 running Tiger using GCC 4.0.1 in 295 minutes.
Built on Tiger/i386 with a 2Ghz c2d mac mini using GCC 4.0.1 in 63 minutes
Built on 1st gen 1.8Ghz iMac G5 running Leopard using GCC 4.2 in 312 minutes.
Built on mid-2009 Air running Leopard using GCC 4.2 in 77 minutes.
Built on mid-2009 Air running Snow Leopard using GCC 4.2 in 99 minutes.
Built on late 2009 white MacBook running Lion using clang in 45 minutes
Built on late 2009 white MacBook running Mountain Lion using clang in 48 minutes
Built on late 2009 white MacBook running Mavericks using clang in 52 minutes
Built on late 2009 white MacBook running Yosemite using clang in 73 minutes
Built on mid-2009 Air running El Capitan using clang in 88 minutes.

Marking as draft until it's been done more builds.

Unable to build on Leopard (i386/powerpc) without setting `no_optimization` & modifying `HOMEBREW_OPTIMIZATION_LEVEL` to be set to `O0` instead of `Os`, otherwise once `xgcc` is built at stage 1, things fail at the configure stage for libgcc, with an `internal compiler error: Bus error` due to "optimisation".
On Tiger powerpc need to set `no_optimization` but didn't have to edit `HOMEBREW_OPTIMIZATION_LEVEL`.

### DIFF
--- a/Library/Formula/gcc47.rb
+++ b/Library/Formula/gcc47.rb
@@ -77,6 +77,12 @@ class Gcc47 < Formula
   patch :DATA
 
   def install
+    # GCC Bug 25127 for PowerPC
+    # https://gcc.gnu.org/bugzilla//show_bug.cgi?id=25127
+    # ../../../libgcc/unwind.inc: In function '_Unwind_RaiseException':
+    # ../../../libgcc/unwind.inc:136:1: internal compiler error: in rs6000_emit_prologue, at config/rs6000/rs6000.c:19445
+    # "internal compiler error: Bus error" on Leopard
+    ENV.no_optimization
     # GCC will suffer build errors if forced to use a particular linker.
     ENV.delete "LD"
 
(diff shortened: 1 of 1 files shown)


Library/Formula/gcc47.rb
```diff
@@ -122,15 +117,8 @@ def install
       # don't wander around telling little children there is no Santa
       # Claus.
       "--enable-version-specific-runtime-libs",
-      "--enable-libstdcxx-time=yes",
```
mistydemeo, line 125: I guess we probably need to revbump it if removing it will also change compat?

mistydemeo (approved):

Library/Formula/gcc47.rb
sevan: Yes, you're right.

PR #1315, opened 2025-03-25
tree: update to 2.2.1

aarossig:
- The old links are long-dead.
- Remove unused prebuilts.
- Update to the latest

Tested: brew install ./tree.rb


sevan:
Thanks, tested on Tiger/PowerPC with GCC 4.0.1.

Issue #1332, opened 2025-04-20
erlang fails to build on Leopard


jcgraybill:
I'm unable to build erlang on a Mac OS X Leopard system (PPC G4) - encountering three errors:

1. Checksum mismatch for the OTP source tarball
2. OTP 18 build incompatibility with version 2.72 of `autoconf`, which is the Tigerbrew version: https://github.com/asdf-vm/asdf-erlang/issues/195
3. OTP 18 build incompatibility with version 1.1 of openssl, which is the Tigerbrew version: https://github.com/asdf-vm/asdf/issues/195

I propose updating the OTP version from 18.1 to 18.1.5, which is the latest patch release for OTP 18.1, and switching to using system-provided versions of both autoconf and openssl. I'm testing a build with this configuration right now. If folks agree, I'll put together a PR.

Thanks!

sevan:
> I'm unable to build erlang on a Mac OS X Leopard system (PPC G4) - encountering three errors:

> 2. OTP 18 build incompatibility with version 2.72 of `autoconf`, which is the Tigerbrew version: [MacOS Big Sur + Erlang 22.3 + AutoConf -> cannot find required auxiliary files: install-sh config.guess config.sub asdf-vm/asdf-erlang#195](https://github.com/asdf-vm/asdf-erlang/issues/195)
  
Looks like for some reason automake is not putting those into place, but no worries on using the OS supplied version of tools. Saves on fighting an old version of software to work with current autotools.
  
> 3. OTP 18 build incompatibility with version 1.1 of openssl, which is the Tigerbrew version: [Unable to build erlang 19.x asdf-vm/asdf#195](https://github.com/asdf-vm/asdf/issues/195)

Right, at a glance (I'm not an Erlang programmer) support for newer releases of OpenSSL don't start landing until version 20 of OTP.

> I propose updating the OTP version from 18.1 to 18.1.5, which is the latest patch release for OTP 18.1, and switching to using system-provided versions of both autoconf and openssl. I'm testing a build with this configuration right now. If folks agree, I'll put together a PR.

Patches are welcome :)
Is there a reason why you want to stick with 18.1.x specifically? (jump to 18.3.4.11 or higher major releases if powerpc support is still there?)

sevan:
> Mostly I was just being conservative with my first PR, and I wasn't sure how much appetite there was for version bumps.

ah, no worries :)

[...]

PR #1333, opened 2025-04-21
maven: fix download url and java dependency

jcgraybill:
Remove URL that's returning 404, promote mirror URL to main URL. Specify that Maven 3.3.3 requires Java 8, so that systems with lower versions fail the dependency check rather than fail with build errors.

Tested package download/dependency check on Leopard/PowerPC, but I don't have Java 8 on my development machine to test the build beyond this.

Resolves a maven build issue identified in https://github.com/mistydemeo/tigerbrew/issues/1332

### DIFF
--- a/Library/Formula/maven.rb
+++ b/Library/Formula/maven.rb
@@ -1,11 +1,10 @@
 class Maven < Formula
   desc "Java-based project management"
   homepage "https://maven.apache.org/"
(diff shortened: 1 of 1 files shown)


sevan:
Thanks.
There is an experimental [build of Java 8 for Darwin/powerpc btw](https://github.com/nilsvanvelzen/mac_ppc_openjdk8u60)

jcgraybill:
Thanks for that tip - taking a look.

PR #1344, opened 2025-04-24
erlang 18.3.4.11

jcgraybill:
Update to the final released version of Erlang 18.

I've tested this on Leopard/PowerPC (G4). All dependent packages build with their current versions, but I haven't yet tested `ejabberd`, `disco`, `discodex`, and `couchdb-lucene`, since these also didn't build correctly under the previous versions of `erlang`. 

As a future project, I'd be interested to explore upgrading to higher major version releases of `erlang`. I briefly tried to build a version of Erlang 19 using this formula, but it didn't work, and will take more investigation. So in the meantime, I think it'd be great to at least get `erlang` and its dependents bumped up to the latest minor/patch releases of their major release versions.

### DIFF
--- a/Library/Formula/erlang.rb
+++ b/Library/Formula/erlang.rb
@@ -5,22 +5,22 @@ class Erlang < Formula
   desc "Erlang Programming Language"
   homepage "http://www.erlang.org"
 
-  stable do
-    # Download tarball from GitHub; it is served faster than the official tarball.
-    url "https://github.com/erlang/otp/archive/OTP-18.1.5.tar.gz"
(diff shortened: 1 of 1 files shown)


sevan:
Thanks for looking into this. If you ever get the chance and find in the end it's not possible to move beyond version 18, maybe we can look at bring back support for OpenSSL beyond 1.1.0 in `crypto.c` to 18, so it can be linked to current formula.

sevan:
Built on Tiger (i386/G5) with GCC 4.0.1.

67 minutes on the 1.8Ghz iMac G5
27 minutes on the 2Ghz c2d mac mini

jcgraybill:
Thanks, sevan. I'll take a look into OpenSSL support in parallel with looking at higher major versions.

PR #1346, opened 2025-04-24
elixir 1.5.3

jcgraybill:
Update elixir to 1.5.3 and remove outdated bottle checksums. This is the latest version of `elixir` that is compatible with Erlang 18. Tested both with the current version of `erlang` as well as the version in PR https://github.com/mistydemeo/tigerbrew/pull/1344.


sevan:
Thanks.

PR #1347, opened 2025-04-24
wrangler 1.2

jcgraybill:
Update wrangler to its latest release, 1.2. Tested with both the current version of `erlang` and https://github.com/mistydemeo/tigerbrew/pull/1344.


sevan:
Tested on Tiger (i386/G5) with GCC 4.0.1

PR #1348, opened 2025-04-24
syntaxerl 0.13.0

jcgraybill:
Update syntaxerl to version 0.13.0, later versions do not produce a `syntaxerl` program in /usr/local/bin/ which seems to be due to a `rebar3` dependency, which itself depends on OTP 19. No release notes are provided in Github for syntaxerl, so this may be resolvable under OTP 18, but will require more investigation. Tested against https://github.com/mistydemeo/tigerbrew/pull/1344.


jcgraybill:
Indeed, those tests were wrong. I've fixed them and tested both positive and negative cases for both. Thanks for catching that!

sevan:
Thanks for the fix. :)

PR #1350, opened 2025-04-24
disco fix install issues

jcgraybill:
It looks like the formula for disco hasn't worked for quite some time: this commit fixes an issue where rebar tries to fetch dependencies using deprecated URLs, as well as two very small OTP incompatibilities.

Tested on Leopard/PPC (G4). The one dependent package (`discodex`) builds correctly now too, and doesn't require any changes.

### DIFF
--- a/Library/Formula/disco.rb
+++ b/Library/Formula/disco.rb
@@ -75,7 +75,7 @@ class Disco < Formula
 
     # Fix the config file to point at the linked files, not in to cellar
     # This isn't ideal - if there's a settings.py file left over from a previous disco
-    # installation, it'll issue a Warning
+    # installation, it'll issue an error
     inreplace "#{etc}/disco/settings.py" do |s|
       s.gsub!("Cellar/disco/"+version+"/", "")
     end
--- a/Library/Formula/disco.rb
+++ b/Library/Formula/disco.rb
@@ -28,9 +28,24 @@ class Disco < Formula
       s.change_make_var! "localstatedir", var
     end
 
+    # Ignore warnings about erlang:now() deprecation in OTP 18
+    inreplace "master/rebar.config", "warnings_as_errors,", ""
+
     # Disco's "rebar" build tool refuses to build unless it's in a git repo, so
     # make a dummy one
     system "git init && git add master/rebar && git commit -a -m 'dummy commit'"
+
+    # rebar tries to clone goldrush, bear, and meck using git://github.com/ urls
+    # which are no longer offered by github. Clone these dependencies to the expected
+    # place using https.
+    mkdir_p "master/deps"
+    system "git", "clone", "--branch", "0.1.6", "https://github.com/DeadZen/goldrush.git", "master/deps/goldrush"
+    system "git", "clone", "--branch", "0.8.1", "https://github.com/boundary/bear.git", "master/deps/bear"
+    system "git", "clone", "--branch", "0.8.2", "https://github.com/eproxus/meck", "master/deps/meck"
+
+    # fix erlang 18 incompatibility in type name
+    inreplace "master/src/worker_throttle.erl", "-opaque state() :: queue().", "-opaque state() :: queue:queue()."
+    inreplace "master/src/worker_throttle.erl", "-spec throttle(queue()", "-spec throttle(queue:queue()"
 
     system "make"
     system "make", "install"
@@ -46,16 +61,16 @@ class Disco < Formula
     bin.env_script_all_files(libexec+"bin", :PYTHONPATH => ENV["PYTHONPATH"])
   end
 
-  test do
-    system "#{bin}/disco"
-  end
-
   def caveats
     <<-EOS.undent
       Please copy #{etc}/disco/settings.py to ~/.disco and edit it if necessary.
       The DDFS_*_REPLICA settings have been set to 1 assuming a single-machine install.
       Please see http://discoproject.org/doc/disco/start/install.html for further instructions.
     EOS
+  end
+
+  test do
+    system "#{bin}/disco"
   end
 end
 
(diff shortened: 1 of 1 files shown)


sevan:
It fails but completes the installation for me.
After setting symlinks it reports
```
ln -s ../../../Cellar/disco/0.5.4/lib/python2.7/site-packages/disco-0.5.4-py2.7.egg-info disco-0.5.4-py2.7.egg-info
Traceback (most recent call last):
  File "<string>", line 1, in ?
AttributeError: 'module' object has no attribute 'getusersitepackages'
Traceback (most recent call last):
  File "<string>", line 1, in ?
AttributeError: 'module' object has no attribute 'getusersitepackages'
```

jcgraybill:
Is this on a Snow Leopard machine,  by any chance?

Library/Formula/disco.rb
```diff
@@ -13,25 +13,62 @@ class Disco < Formula
 
   depends_on :python if MacOS.version <= :snow_leopard
```
mistydemeo, line 14: Maybe we should promote this to always requiring a Tigerbrew python, given the age of even Snow Leopard's?
```suggestion
  depends_on :python
```

jcgraybill: That's my hypothesis too, based on the content of the error message. I'll test a build tonight and see whether I can reproduce this.

mistydemeo: Thank you!

jcgraybill:
Well, hm, I can't seem to reproduce this failure. I've tried on both Leopard/PowerPC and Snow Leopard/Intel systems. For both, the brew `python` package shows up as a dependency, installs correctly if it isn't already installed, and then installs the `simplejson` resource in the right place.

@sevan would you mind sharing details about what system / configuration you encountered the error on? 

Thanks!

sevan:
I was building on Tiger/powerpc

jcgraybill:
I'm wondering if this is a manifestation of this issue https://github.com/pypa/virtualenv/issues/804 about brew not playing nice with Python virtual environments. I've been able to recreate the error message in a plain python interpreter, but I'm not able to track down how the `disco` install would encounter that yet. :(

sevan:
We're not doing anything with `virtualenv` though otherwise it'd be consistent across all OS X versions?

jcgraybill:
What a puzzle. Okay, I'll keep trying to reproduce this build failure.

sevan:
Had a quick look at the formula, noticed this comment in there
```
# Disco's "rebar" build tool refuses to build unless it's in a git repo, so
# make a dummy one
```

So I added `git` as a build dependency and the build failed in a new way :D

```
cp -r master/www /opt/tigerbrew/Cellar/disco/0.5.4/share/disco/master                                             
/opt/tigerbrew/Library/Homebrew/debrew.rb:11:in `raise'                                                           
Utils::InreplaceError: inreplace failed                                                                                                           
/opt/tigerbrew/etc/disco/settings.py:                                                                             
  expected replacement of "Cellar/disco/0.5.4/" with ""
```

I suspect the fallout of is from updated components with Python 2 as the Python documentation says `getusersitepackages` was [introduced in Python 3.2](https://docs.python.org/3/library/site.html#site.getusersitepackages).

Worth switching to Python 3 anyway? judging by a couple of pull requests in the [project's repo](https://github.com/discoproject/disco) people are running it on Python 3?

jcgraybill:
Oh good, I _do_ know all about that new error. This happens when you install, remove, and then reinstall the package, because it leaves behind one settings file. There's a note in the formula about this. Basically, you need to manually `rm /opt/tigerbrew/etc/disco/settings.py` after removing the `disco` package.

```
    # Fix the config file to point at the linked files, not in to cellar
    # This isn't ideal - if there's a settings.py file left over from a previous disco
    # installation, it'll issue an error
```
I agree it "isn't ideal". I think it would be much nicer for users to either remove the settings file when removing the package. Or if we would want to leave the settings file in place (so people keep any customized settings they create) to gracefully fail or skip the `inreplace` command.

jcgraybill:
Mixed news: `disco` builds with the Python version changed to `python3`, but it won't launch. There's quite a bit of reference to Python 2-specific libraries and paths in the code, which people appear to have been hacking around in their local forks to keep it working. I've reverted to `python` 2 in the PR for now, but that leaves https://github.com/mistydemeo/tigerbrew/pull/1350#issuecomment-2849481472 problem active, so the PR isn't in condition to merge yet.

I'm a little worried to see how many PRs and issues report incompatibilities with versions of Erlang/OTP later than 18. I'm separately exploring whether it'd be possible to [upgrade the `erlang` version in Tigerbrew](https://github.com/jcgraybill/tigerbrew/issues/1). There are early signs that a more recent Erlang/OTP version would bring a lot of benefit to packages/products that are being actively maintained and probably used more than `disco`.

I wonder if we should pause this effort, and decide whether `disco` is worth fixing after I have more interesting thoughts to offer about an Erlang/OTP upgrade.

sevan:
> Mixed news: `disco` builds with the Python version changed to `python3`, but it won't launch. There's quite a bit of reference to Python 2-specific libraries and paths in the code, which people appear to have been hacking around in their local forks to keep it working. I've reverted to `python` 2 in the PR for now, but that leaves [#1350 (comment)](https://github.com/mistydemeo/tigerbrew/pull/1350#issuecomment-2849481472) problem active, so the PR isn't in condition to merge yet.

Ok.

> I'm a little worried to see how many PRs and issues report incompatibilities with versions of Erlang/OTP later than 18. I'm separately exploring whether it'd be possible to [upgrade the `erlang` version in Tigerbrew](https://github.com/jcgraybill/tigerbrew/issues/1). There are early signs that a more recent Erlang/OTP version would bring a lot of benefit to packages/products that are being actively maintained and probably used more than `disco`.

Oh, cool. From your table, OTP 21 & 22 looks promising?

> I wonder if we should pause this effort, and decide whether `disco` is worth fixing after I have more interesting thoughts to offer about an Erlang/OTP upgrade.

Sure

jcgraybill:
Very excited to get a version of Erlang with openssl support working.

Closing for now. Thank you @sevan and @mistydemeo !

PR #1352, opened 2025-04-25
ejabberd 16.12

jcgraybill:
Updating ejabberd to 16.12. Version 16 is the first version I was able to build on Leopard/PPC, and 16.12 is the highest minor version of ejabberd 16.

Also switch to using openssl3.

Build successfully tested on Leopard/PPC, and sure enough I can connect Adium clients to the server. I would love help testing on Snow Leopard if anyone is able to, as the pam_appl.h workaround I needed to add is apparently not needed on SL, but I don't currently have a way to test that.

### DIFF
--- a/Library/Formula/ejabberd.rb
+++ b/Library/Formula/ejabberd.rb
@@ -1,8 +1,8 @@
 class Ejabberd < Formula
   desc "XMPP application server"
   homepage "https://www.ejabberd.im"
-  url "https://www.process-one.net/downloads/ejabberd/15.07/ejabberd-15.07.tgz"
-  sha256 "87d5001521cbb779b84bc74879e032e2514d9a651e24c4e40cce0907ab405bd1"
+  url "https://github.com/processone/ejabberd/archive/refs/tags/15.11.tar.gz"
+  sha256 "75dcb533c04df5926cc118ca997c843ff5a8a80ef0b49782d860e9d7cbe76ca0"
 
   head "https://github.com/processone/ejabberd.git"
 
@@ -38,6 +38,7 @@ class Ejabberd < Formula
             "--enable-odbc",
             "--enable-pam"]
 
+    system "./autogen.sh"
     system "./configure", *args
     system "make"
     system "make", "install"
--- a/Library/Formula/ejabberd.rb
+++ b/Library/Formula/ejabberd.rb
@@ -1,14 +1,12 @@
 class Ejabberd < Formula
   desc "XMPP application server"
   homepage "https://www.ejabberd.im"
-  url "https://github.com/processone/ejabberd/archive/refs/tags/15.11.tar.gz"
-  sha256 "75dcb533c04df5926cc118ca997c843ff5a8a80ef0b49782d860e9d7cbe76ca0"
-
-  head "https://github.com/processone/ejabberd.git"
+  url "https://github.com/processone/ejabberd/archive/refs/tags/16.12.tar.gz"
+  sha256 "a7eeb9fe49ef141daab1be01838a7612dff9194a28c3dfc922cc691bb8c9b532"
 
   option "32-bit"
 
-  depends_on "openssl"
+  depends_on "openssl3"
   depends_on "erlang"
   depends_on "libyaml"
   # for CAPTCHA challenges
@@ -29,12 +27,26 @@ class Ejabberd < Formula
             "--enable-pgsql",
             "--enable-mysql",
             "--enable-odbc",
-#            "--enable-pam"
+            "--enable-pam"
           ]
 
+    system "autoupdate"
     system "./autogen.sh"
     system "./configure", *args
-    system "git", "clone", "--branch", "0.1.7", "https://github.com/DeadZen/goldrush.git", "deps/goldrush"
+
+    # rebar tries to clone goldrush using a git://github.com/ url which is no longer offered by github.
+    # Work around by cloning it into the expected place using https.
+    system "git", "clone", "--branch", "0.1.8", "https://github.com/DeadZen/goldrush.git", "deps/goldrush"
+
+    # Before Snow Leopard, the pam header files were in /usr/include/pam instead of /usr/include/security.
+    # https://trac.macports.org/ticket/26127
+    if MacOS.version <= :leopard
+      system "git", "clone", "--branch", "1.0.0", "https://github.com/processone/epam", "deps/p1_pam"
+      inreplace "deps/p1_pam/configure", "security/pam_appl.h", "pam/pam_appl.h"
+      inreplace "deps/p1_pam/configure.ac", "security/pam_appl.h", "pam/pam_appl.h"
+      inreplace "deps/p1_pam/c_src/epam.c", "security/pam_appl.h", "pam/pam_appl.h"
+    end
+
     system "make"
     system "make", "install"
 
--- a/Library/Formula/ejabberd.rb
+++ b/Library/Formula/ejabberd.rb
@@ -6,13 +6,6 @@ class Ejabberd < Formula
 
   head "https://github.com/processone/ejabberd.git"
 
-  bottle do
-    sha256 "4001c8bf43972697862ead67f944177c6a6b771c96a4d6de75f68694ed7aa620" => :el_capitan
-    sha256 "592a3412890d52da9d8a9f43a288dca8eab233dcd0c59d17d4bd8f89b7b4567b" => :yosemite
-    sha256 "17b97c88ea724e8816c445a944ad71a9be141cf9afe8e4f168dd271ea2bd8448" => :mavericks
-    sha256 "f53c7307acaee6aafa813785ede998900c0c32db2d0f28dfb5dd51c45b6f5366" => :mountain_lion
-  end
-
   option "32-bit"
 
   depends_on "openssl"
@@ -36,10 +29,12 @@ class Ejabberd < Formula
             "--enable-pgsql",
             "--enable-mysql",
             "--enable-odbc",
-            "--enable-pam"]
+#            "--enable-pam"
+          ]
 
     system "./autogen.sh"
     system "./configure", *args
+    system "git", "clone", "--branch", "0.1.7", "https://github.com/DeadZen/goldrush.git", "deps/goldrush"
     system "make"
     system "make", "install"
 
--- a/Library/Formula/ejabberd.rb
+++ b/Library/Formula/ejabberd.rb
@@ -6,9 +6,14 @@ class Ejabberd < Formula
 
   option "32-bit"
 
-  depends_on "openssl3"
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "rebar" => :build
+
   depends_on "erlang"
   depends_on "libyaml"
+  depends_on "openssl3"
+
   # for CAPTCHA challenges
   depends_on "imagemagick" => :optional
 
@@ -27,16 +32,13 @@ class Ejabberd < Formula
             "--enable-pgsql",
             "--enable-mysql",
             "--enable-odbc",
-            "--enable-pam"
-          ]
+            "--enable-pam"]
+
+    inreplace "rebar.config", 'lager", {tag, "3.2.1', 'lager", {tag, "3.2.4'
 
     system "autoupdate"
     system "./autogen.sh"
     system "./configure", *args
-
-    # rebar tries to clone goldrush using a git://github.com/ url which is no longer offered by github.
-    # Work around by cloning it into the expected place using https.
-    system "git", "clone", "--branch", "0.1.8", "https://github.com/DeadZen/goldrush.git", "deps/goldrush"
 
     # Before Snow Leopard, the pam header files were in /usr/include/pam instead of /usr/include/security.
     # https://trac.macports.org/ticket/26127


sevan:
When attempting to link against openssl or openssl3, it fails for me on Tiger.
```
Compiling /private/tmp/ejabberd20250425-20607-m5p9hm/ejabberd-16.12/deps/fast_tls/c_src/fast_tls_drv.c
/private/tmp/ejabberd20250425-20607-m5p9hm/ejabberd-16.12/deps/fast_tls/c_src/fast_tls_drv.c: In function ‘setup_ecdh’:
/private/tmp/ejabberd20250425-20607-m5p9hm/ejabberd-16.12/deps/fast_tls/c_src/fast_tls_drv.c:386: error: ‘EC_KEY’ undeclared (first use in this function)
/private/tmp/ejabberd20250425-20607-m5p9hm/ejabberd-16.12/deps/fast_tls/c_src/fast_tls_drv.c:386: error: (Each undeclared identifier is reported only once
/private/tmp/ejabberd20250425-20607-m5p9hm/ejabberd-16.12/deps/fast_tls/c_src/fast_tls_drv.c:386: error: for each function it appears in.)
/private/tmp/ejabberd20250425-20607-m5p9hm/ejabberd-16.12/deps/fast_tls/c_src/fast_tls_drv.c:386: error: ‘ecdh’ undeclared (first use in this function)
/private/tmp/ejabberd20250425-20607-m5p9hm/ejabberd-16.12/deps/fast_tls/c_src/fast_tls_drv.c:393: error: ‘SSL_OP_SINGLE_ECDH_USE’ undeclared (first use in this function)
ERROR: compile failed while processing /private/tmp/ejabberd20250425-20607-m5p9hm/ejabberd-16.12/deps/fast_tls: rebar_abort
```

You can skip cloning the git repos manually, but patching the `rebar.config` to switch the urls from `git://` protocol to `https://`. Need to add `rebar` as a build dep.

e.g.
```
--- rebar.config.orig   2025-04-25 13:51:37.000000000 +0100
+++ rebar.config        2025-04-25 13:51:50.000000000 +0100
@@ -27,7 +27,7 @@
 {port_specs, [{"priv/lib/fast_tls_drv.so", ["c_src/fast_tls_drv.c"]},
               {"priv/lib/p1_sha.so", ["c_src/p1_sha.c"]}]}.
 
-{deps, [{p1_utils, ".*", {git, "git://github.com/processone/p1_utils", {tag, "1.0.6"}}}]}.
+{deps, [{p1_utils, ".*", {git, "https://github.com/processone/p1_utils", {tag, "1.0.6"}}}]}.
 
 {clean_files, ["c_src/fast_tls_drv.gcda", "c_src/fast_tls_drv.gcno",
                "c_src/fast_sha.gcda", "c_src/fast_sha.gcno"]}.
```

One for me as a reminder:
Since the original autoconf'd source archives have disappeared and it requires to run autogen now, best to add autoconf & automake as build deps since the versions in Tiger are too old.

sevan:
Ok, regarding cloning repos manually, the problem is that there's a bunch of deps which are listed in the `rebar.config` included in the root of the source for ejabberd. These deps use HTTPS urls, unfortunately the deps themselves include `rebar.config` files which list their own dependencies and there is where the git protocol URLs are showing up. For the `goldrush`dep, it is introduce via the Basho fork of lager, upping the version fixed the issue as they switched to HTTPS for a fork of goldrush made by Basho.

```
--- rebar.config.orig   2025-04-25 14:32:33.000000000 +0100
+++ rebar.config        2025-04-25 14:33:42.000000000 +0100
@@ -7,7 +7,7 @@
 %%% Created :  1 May 2013 by Evgeniy Khramtsov <ekhramtsov@process-one.net>
 %%%-------------------------------------------------------------------

-{deps, [{lager, ".*", {git, "https://github.com/basho/lager", {tag, "3.2.1"}}},
+{deps, [{lager, ".*", {git, "https://github.com/basho/lager", {tag, "3.2.4"}}},
         {p1_utils, ".*", {git, "https://github.com/processone/p1_utils", {tag, "1.0.6"}}},
         {cache_tab, ".*", {git, "https://github.com/processone/cache_tab", {tag, "1.0.5"}}},
         {fast_tls, ".*", {git, "https://github.com/processone/fast_tls", {tag, "1.0.9"}}},
```

Haven't make any progress on the OpenSSL build issue.

jcgraybill:
> When attempting to link against openssl or openssl3, it fails for me on Tiger.

Oh, interesting! Between version `15.07` and `16.12`, `ejabberd` switched from using their [tls](https://github.com/processone/tls) library to a new one called [fast_tls](https://github.com/processone/fast_tls). I couldn't get `tls` to build on Leopard, but I wonder if `fast_tls` isn't backward compatible with Tiger. 

It would be interesting to see whether it would be possible to hack `tls` back into `ejabberd` 16.12 and build it for Tiger. I don't actually know whether these two libraries are API-compatible. If that doesn't work, this might be a tough situation where there isn't a single version of `ejabberd` that builds on all three of Tiger, Leopard, and Snow Leopard. 

In a case like that, what approach do you all like to take in this project? My mind is going to, is there some way to provide different software versions based on OS version? I haven't seen a formula that does this, but I've only read a very small percentage of the tigerbrew formulas. Other ideas could be to try cross-compiling and releasing a bottle for Tiger users, or to somehow mark the package as incompatible with Tiger. Or of course, to keep hacking at it until I get something to work across all three :)

I think the moral is, I really need to get Tiger and Snow Leopard onto some machines here so that I can test all of the target OS versions myself. There are a couple Mac Minis in the mail right now that should take care of that shortly.

> You can skip cloning the git repos manually, but patching the `rebar.config` to switch the urls from `git://` protocol to `https://`. 

Oh, thanks for that tip. I've been trying to use `inreplace` whenever possible, too. The challenge I found with `ejabberd` is that `goldrush` is a transitive dependency, via `lager`, so the URL that needs to be fixed is in `lager`'s own `rebar.config`. That file doesn't exist in the build directory until `rebar compile` has already been called - at which point rebar downloads `lager`, and then tries to use `lager`'s `rebar.config` to fetch `goldbar` and fails. I don't know if there's a way to return control to tigerbrew in the midst of this process so that I can apply the needed patches. However, if `rebar` finds `goldrush` already downloaded, it skips trying to download it again. You'll see that I had to do this with three erlang libraries in `disco.rb`. It was also the only way I could figure out to patch the include paths in `epam`.

> Need to add `rebar` as a build dep.

> One for me as a reminder: Since the original autoconf'd source archives have disappeared and it requires to run autogen now, best to add autoconf & automake as build deps since the versions in Tiger are too old.

I'll go ahead and add all three of these to the branch in this PR and do a test build tonight or this weekend. (`erlang` is compiling right now, and wow does it take a long time.) True confession, I also realize I haven't tried a build that includes `imagemagick` and that might be a whole can of worms too. 

Thanks as always for all the help, @sevan .

jcgraybill:
> Ok, regarding cloning repos manually, the problem is that there's a bunch of deps which are listed in the `rebar.config` included in the root of the source for ejabberd. These deps use HTTPS urls, unfortunately the deps themselves include `rebar.config` files which list their own dependencies and there is where the git protocol URLs are showing up. For the `goldrush`dep, it is introduce via the Basho fork of lager, upping the version fixed the issue as they switched to HTTPS for a fork of goldrush made by Basho.

Haha, looks like we were typing at the same time. This is a great find - it's not a very big version change for `lager`, so it looks like a more robust way to solve this.

sevan:
> > When attempting to link against openssl or openssl3, it fails for me on Tiger.
> 
> Oh, interesting! Between version `15.07` and `16.12`, `ejabberd` switched from using their [tls](https://github.com/processone/tls) library to a new one called [fast_tls](https://github.com/processone/fast_tls). I couldn't get `tls` to build on Leopard, but I wonder if `fast_tls` isn't backward compatible with Tiger.
> 
> It would be interesting to see whether it would be possible to hack `tls` back into `ejabberd` 16.12 and build it for Tiger. I don't actually know whether these two libraries are API-compatible. If that doesn't work, this might be a tough situation where there isn't a single version of `ejabberd` that builds on all three of Tiger, Leopard, and Snow Leopard.

In this case, I suspect the issues is not with Tiger itself, but it is being highlighted there since regardless of OS, we're trying to link to the `openssl3` formula (also tried `openssl` without success). I don't see anything about flags for pointing to a TLS library in the `./configure --help` output

> In a case like that, what approach do you all like to take in this project? My mind is going to, is there some way to provide different software versions based on OS version? I haven't seen a formula that does this, but I've only read a very small percentage of the tigerbrew formulas. Other ideas could be to try cross-compiling and releasing a bottle for Tiger users, or to somehow mark the package as incompatible with Tiger. Or of course, to keep hacking at it until I get something to work across all three :)

I did this for [Freepascal](https://github.com/mistydemeo/tigerbrew/pull/1164) which is not ready yet.
I think in this scenario, I suspect there's a bug in the build infrastructure because I have a current version of OpenSSL installed (1.1.0 and 3.50).  

> I think the moral is, I really need to get Tiger and Snow Leopard onto some machines here so that I can test all of the target OS versions myself. There are a couple Mac Minis in the mail right now that should take care of that shortly.

More machines running Tiger are always good :)
I can try testing on Snow Leopard over the weekend.


> Oh, thanks for that tip. I've been trying to use `inreplace` whenever possible, too. The challenge I found with `ejabberd` is that `goldrush` is a transitive dependency, via `lager`, so the URL that needs to be fixed is in `lager`'s own `rebar.config`. That file doesn't exist in the build directory until `rebar compile` has already been called - at which point rebar downloads `lager`, and then tries to use `lager`'s `rebar.config` to fetch `goldbar` and fails. I don't know if there's a way to return control to tigerbrew in the midst of this process so that I can apply the needed patches. However, if `rebar` finds `goldrush` already downloaded, it skips trying to download it again. You'll see that I had to do this with three erlang libraries in `disco.rb`. It was also the only way I could figure out to patch the include paths in `epam`.

Yeah, I was sitting in the build directory, having spawned a shell when the build failed and generated the patches then, only to relies that the deps directory is not there at the start and the patching fails.


> I'll go ahead and add all three of these to the branch in this PR and do a test build tonight or this weekend. (`erlang` is compiling right now, and wow does it take a long time.) True confession, I also realize I haven't tried a build that includes `imagemagick` and that might be a whole can of worms too.

Oh, and all the security issues with imagemagick :D

> Thanks as always for all the help, @sevan .

Thank you for all your contributions so far @jcgraybill. It's cool to have a version of erlang to play with. LFE looks very interesting. :)

Library/Formula/ejabberd.rb
```diff
@@ -36,9 +27,26 @@ def install
...
+    system "./autogen.sh"
     system "./configure", *args
+
+    # rebar tries to clone goldrush using a git://github.com/ url which is no longer offered by github.
+    # Work around by cloning it into the expected place using https.
+    system "git", "clone", "--branch", "0.1.8", "https://github.com/DeadZen/goldrush.git", "deps/goldrush"
+
+    # Before Snow Leopard, the pam header files were in /usr/include/pam instead of /usr/include/security.
+    # https://trac.macports.org/ticket/26127
+    if MacOS.version <= :leopard
```
sevan, line 43: Comment mentions Snow Leopard

sevan:
As it currently stands, what does `otool -L opt/ejabberd/lib/fast_tls-1.0.9/priv/lib/fast_tls_drv.so` and `otool -L opt/ejabberd/lib/fast_tls-1.0.9/priv/lib/p1_sha.so` say on your system?
They should have linked to the `openssl3` formula's libraries.

jcgraybill:
It looks like they did link `openssl3`:

```
kabooties-emac:local kabootie$ otool -L opt/ejabberd/lib/fast_tls-1.0.9/priv/lib/fast_tls_drv.so
opt/ejabberd/lib/fast_tls-1.0.9/priv/lib/fast_tls_drv.so:
	/usr/local/opt/openssl3/lib/libssl.3.dylib (compatibility version 3.0.0, current version 3.0.0)
	/usr/local/opt/openssl3/lib/libcrypto.3.dylib (compatibility version 3.0.0, current version 3.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 111.1.7)
```

```
kabooties-emac:local kabootie$ otool -L opt/ejabberd/lib/fast_tls-1.0.9/priv/lib/p1_sha.so
opt/ejabberd/lib/fast_tls-1.0.9/priv/lib/p1_sha.so:
	/usr/local/opt/openssl3/lib/libssl.3.dylib (compatibility version 3.0.0, current version 3.0.0)
	/usr/local/opt/openssl3/lib/libcrypto.3.dylib (compatibility version 3.0.0, current version 3.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 111.1.7)
```

Library/Formula/ejabberd.rb
jcgraybill: Yeah, that comment in MacPorts is a little ambiguous to me. The way I ended up interpreting it is that `pam_appl.h` locations are:

- tiger: `/usr/include/pam`
- leopard: `/usr/include/pam`
- snow leopard: `/usr/include/security`

sevan: My mistake, appologies.
I misread it as "< :leopard", while the comment says "before snow leopard".

sevan:
> It looks like they did link `openssl3`:

How weird, not for me.
I'll take a look at testing on other OS X versions in the morning and report back if it has issues on them but I'll land this now in the meantime.

Thanks again for patches.

jcgraybill:
That *is* weird. I'll definitely be standing by to look into anything you find.

Also, coda: the build succeeded on Leopard/PPC with the `--with-imagemagick` flag, and I'm currently test-building the 32-bit build. 

Thanks!

sevan:
Build tested on OS X 10.5 i386 & 10.6 x86_64.
`erlang` needed a change to build on 10.6, but otherwise everything build without issue.
On both version OS, `fast_tls` library linked to `openssl3` without issue.
I have a hunch what it may be on Tiger but I'll take a look another day.

I was wondering if maybe instead of fetching the deps during build, we could switch to using resources in the formula, that way `brew` will fetch everything, making it easy to integrate patches, and cache those downloads.
Not sure how you control where `brew` puts the resources though as the build expects to find them in `deps/`

jcgraybill:
Using a resource is an interesting idea. It looks like this is the code that handles resources? https://github.com/mistydemeo/tigerbrew/blob/master/Library/Homebrew/resource.rb

I'm not familiar enough with the brew codebase (or experienced enough at reading Ruby) to immediately see a way to achieve "just unzip the zip file to the specified directory" with that, though, TBH.

sevan:
> I'm not familiar enough with the brew codebase (or experienced enough at reading Ruby)

Likewise but from the link you provided, [doesn't target sort of do that](https://github.com/mistydemeo/tigerbrew/blob/3e492b7046b760d519c5676f9af0e63d2df4c5ba/Library/Homebrew/resource.rb#L84)?

jcgraybill:
Indeed! I got a little tripped up expecting `unpack` to behave similarly to `stage`, but I just have to download & verify the zipfile directly in the formula, and it works great.

Here's a PR with this change: https://github.com/mistydemeo/tigerbrew/pull/1355

Let me also go revise the `disco` PR (https://github.com/mistydemeo/tigerbrew/pull/1350/files) to use the same approach.

Much nicer!

PR #1356, opened 2025-04-27
expat 2.7.1

sevan:
Restore $homepage, dropped by mistake in my 2.7.0 update.


sevan:
Already landed.

PR #1358, opened 2025-04-30
ejabberd: Use prefetched p1_pam for all builds

sevan:
No reason why it needs to be isolated to specific OS version.
While here add a description for the lager version bump tweak.
Try to match code to the comment


jcgraybill:
:)

PR #1362, opened 2025-05-03
ld64 add apple-gcc42 build dependency

jcgraybill:
When building `ld64` on Tiger Intel, a `CompilerSelectionError` is raised, which shows the user this error message:
```
To install this formula, you may need to either:
   brew install apple-gcc42
or:
   brew install gcc
```

Only `apple-gcc42` is a valid compiler to choose for `ld64`: `gcc` is version 8.5.0 and `ld64` is itself one of `gcc`'s dependencies. Adding a build dependency for `apple-gcc42` for Tiger Intel builds. 

Tested on Tiger Intel.

### DIFF
--- a/Library/Formula/ld64.rb
+++ b/Library/Formula/ld64.rb
@@ -26,7 +26,7 @@ class Ld64 < Formula
   depends_on "cctools-headers" => :build
   depends_on "dyld-headers" => :build
   depends_on "libunwind-headers" => :build
-  depends_on "apple-gcc42" => :build if MacOS.version == :tiger && Hardware::CPU.type == :intel
+  depends_on "apple-gcc42" => :build if MacOS.version == :tiger 
 
   keg_only :provided_by_osx,
     "ld64 is an updated version of the ld shipped by Apple."
--- a/Library/Formula/ld64.rb
+++ b/Library/Formula/ld64.rb
@@ -26,6 +26,7 @@ class Ld64 < Formula
   depends_on "cctools-headers" => :build
   depends_on "dyld-headers" => :build
   depends_on "libunwind-headers" => :build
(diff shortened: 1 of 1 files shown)


sevan:
hmm, it's a dependency on 10.4, not exclusive to the intel build. It's just that there's a powerpc package available so you only hit it on the Intel build.
The compiler would only be a build dependency and we don't rely on a compiler package.
The message you're seeing is from `class CompilerSelectionError` (line 356) in `Library/Homebrew/exceptions.rb`.

jcgraybill:
Sure enough, if I build from source on Tiger PowerPC instead of using the bottle, it runs into the same problem.

I realize, the other reason I didn't run into this on PowerPC is that `ld64' is only a dependency for `git` on Tiger/Intel, due to this condition in libffi: https://github.com/mistydemeo/tigerbrew/blob/66dd14b7e808f8fc4d88cb614718636ac89a90a0/Library/Formula/libffi.rb#L15

Now I'm really curious, what's the reason behind the principle of not dependencies to compilers? 

It'd be great for `git` to be able to build without user intervention - or at least for the error message to be more specific. I'm happy to try to tackle this. Is there an approach you think is most preferable?

sevan:
> Now I'm really curious, what's the reason behind the principle of not creating dependencies on compilers?

I guess because you'd still be fighting the compiler selection?
If you have something newer installed, say a more recent version of GCC 4.x the infrastructure would opt to use that?
If you look, there's no formula here which uses `depends_on` for selecting a compiler.

jcgraybill:
Makes sense, I suppose somebody could already have a compiler they like, and wouldn't want it changed. Thanks for the context - seems best to leave this as-is.

PR #1372, opened 2025-05-20
apachetop: update urls, fix formula

jcgraybill:
The `apachetop` source tarballs and homepage are all returning 404. It looks like Homebrew is using source from [this Gitub repository](https://github.com/tessus/apachetop) for `apachetop` now. 

Updating the formula to use the same source and version, removing old bottle hashes, adding dependencies and head url, removing the now-unneeded patch. 

Tested on:

- Tiger PowerPC & Intel
- Leopard PowerPC & Intel
- Snow Leopard Intel


jcgraybill:
FYI this is the pair of lines of code that are being patched in the formula, to confirm that the change made by the patch has already been applied to them. https://github.com/tessus/apachetop/blob/0.23.2/src/resolver.h#L13

Git blame shows they were fixed by a commit in April 2017.

sevan:
How come you need the autotools build deps?

jcgraybill:
Closing this in favor of https://github.com/mistydemeo/tigerbrew/pull/1380

PR #1373, opened 2025-05-20
anttweakbar: remove unused bottle hashes, fix crlf in patch

jcgraybill:
The lines in the patch section of the `anttweakbar` formula end in CRLF, but don't need to. I've test-built this package (Snow Leopard / Intel) with and without those line endings to confirm.

Cleaning this up so the file is easier to work with, and removing unused bottle hashes while I'm at it.


jcgraybill:
This implements a change that I reverted from https://github.com/mistydemeo/tigerbrew/pull/1365, because it's super tricky to make one-line changes to that version of anttweakbar.rb with all of the CRLFs in the patch.

PR #1375, opened 2025-05-21
don't use bashdefault option on bash version <= 2

jcgraybill:
The bash autocomplete script at `Library/Contributions/brew_bash_completion.sh` gives an error when running on the default bash version (2.05b.0(1)-release) on Mac OS X 104 / Tiger: `-bash: complete: bashdefault: invalid option name`. 

Completions work on Tiger if I remove the `bashdefault` option from the `complete` command.  The script works as-is with the default bash version in Leopard (3.2.17(1)-release) and Snow Leopard (3.2.48(1)-release). 

I'm adding a test for which version of bash is in use, and using the appropriate command for bash <= 2 and bash => 3.

### DIFF
--- a/Library/Contributions/brew_bash_completion.sh
+++ b/Library/Contributions/brew_bash_completion.sh
@@ -1,19 +1,9 @@
 # Bash completion script for brew(1)
 #
-# To use, add the following to your .bashrc:
+# To use, install the bash-completion package. Alternately, can be used with 
+# older versions of bash by adding the following to your .bashrc:
 #
 #    . $(brew --repository)/Library/Contributions/brew_bash_completion.sh
-#
-# Alternatively, if you have installed the bash-completion package,
-# you can create a symlink to this file in one of the following directories:
-#
-#    $(brew --prefix)/etc/bash_completion.d
-#    $(brew --prefix)/share/bash-completion/completions
-#
-# Installing to etc/bash_completion.d will cause bash-completion to load
-# it automatically at shell startup time. If you choose to install it to
-# share/bash-completion/completions, it will be loaded on-demand (i.e. the
-# first time you invoke the `brew` command in a shell session).
 
 __brewcomp_words_include ()
 {
--- a/Library/Contributions/brew_bash_completion.sh
+++ b/Library/Contributions/brew_bash_completion.sh
@@ -652,7 +652,7 @@ _brew_to_completion ()
     _brew
 }
 
-if [ `echo "${BASH_VERSION}" | sed "s/\([0-9]*\).*/\1/"` -le 2 ]; then 
+if [ $BASH_VERSINFO -le 2 ]; then 
     complete -o default -F _brew brew
 else
     complete -o bashdefault -o default -F _brew brew
--- a/Library/Contributions/brew_bash_completion.sh
+++ b/Library/Contributions/brew_bash_completion.sh
@@ -72,7 +72,7 @@ __brew_complete_formulae ()
 __brew_complete_installed ()
 {
     local cur="${COMP_WORDS[COMP_CWORD]}"
-    local inst=$(\ls $(brew --cellar))
+    local inst="$(command ls "$(brew --cellar)" 2>/dev/null)"
     COMPREPLY=($(compgen -W "$inst" -- "$cur"))
 }
 
@@ -94,7 +94,7 @@ __brew_complete_versions ()
 __brew_complete_logs ()
 {
     local cur="${COMP_WORDS[COMP_CWORD]}"
-    local logs=$(ls ${HOMEBREW_LOGS:-~/Library/Logs/Homebrew/})
+    local logs="$(command ls "${HOMEBREW_LOGS:-${HOME}/Library/Logs/Homebrew/}" 2>/dev/null)"
     COMPREPLY=($(compgen -W "$logs" -- "$cur"))
 }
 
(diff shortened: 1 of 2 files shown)


sevan:
In the bash-completion formula [I added a heads up about newer bash for Tiger](https://github.com/mistydemeo/tigerbrew/blob/16b634adfc544d1be2259308b42886afe04a02f2/Library/Formula/bash-completion.rb#L12), so there's that option too.

jcgraybill:
Oh, awesome. I think with this change in place too, all our bases will be covered. 

I chose to test for the version of bash rather than the version of Mac OS X, just in case somebody installed a newer bash on an older OS. I'm glad I did that now :)

sevan:
> GNU documentation for bash programmable completions is here. https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html

URL covers current version of bash which supports more completion options.

From `bash(1)` on Tiger:
```
              -o comp-option
                      The comp-option controls several aspects  of  the  comp-
                      spec's  behavior beyond the simple generation of comple-
                      tions.  comp-option may be one of:
                      default Use readline's default  filename  completion  if
                              the compspec generates no matches.
                      dirnames
                              Perform  directory  name completion if the comp-
                              spec generates no matches.
                      filenames
                              Tell readline that the compspec generates  file-
                              names,  so  it can perform any filename-specific
                              processing (like adding  a  slash  to  directory
                              names or suppressing trailing spaces).  Intended
                              to be used with shell functions.
                      nospace Tell  readline  not  to  append  a  space   (the
                              default)  to  words  completed at the end of the
                              line.
```

From `bash(1)` on Leopard:

```
              -o comp-option
                      The  comp-option  controls  several aspects of the comp-
                      spec's behavior beyond the simple generation of  comple-
                      tions.  comp-option may be one of:
                      bashdefault
                              Perform the rest of the default bash completions
                              if the compspec generates no matches.
                      default Use readline's default  filename  completion  if
                              the compspec generates no matches.
                      dirnames
                              Perform  directory  name completion if the comp-
                              spec generates no matches.
                      filenames
                              Tell readline that the compspec generates  file-
                              names,  so  it can perform any filename-specific
                              processing (like adding  a  slash  to  directory
                              names or suppressing trailing spaces).  Intended
                              to be used with shell functions.
                      nospace Tell  readline  not  to  append  a  space   (the
                              default)  to  words  completed at the end of the
                              line.
                      plusdirs
                              After any matches defined by  the  compspec  are
                              generated,    directory   name   completion   is
                              attempted and  any  matches  are  added  to  the
                              results of the other actions.
```

sevan:
Looks like `bashdefault` landed in v3.0.
From its `CWRU/changelog`:
```
pcomplete.h
        - new COPT_BASHDEFAULT and COPT_PLUSDIRS defines

bashline.c
        - if the COPT_BASHDEFAULT flag is set, try the rest of the bash 
          default completions if a compspec with that flag set generates
          no matches
        - broke bash completions out into separate function:
          bash_default_completion(const char *text, int start, int end, 
          int qc, int in_command_position); change attempt_shell_completion
          to call it

bashline.h
        - new extern declaration for bash_default_completion

builtins/complete.def
        - added code to compgen to call bash_default_completion if a compspec
          with the COPT_BASHDEFAULT flag set generates no matches from the
          programmable completion options
```

sevan:
The version of the file currently in Tigerbrew was last updated in May 5th, 2015.
[Development continued upstream](https://github.com/Homebrew/brew/commits/edf000e4cd30c3626ccc28c52ed32f2d84a200dd/Library/Contributions/brew_bash_completion.sh?browsing_rename_history=true&new_path=completions/bash/brew&original_branch=master), and [the file changed location](https://github.com/Homebrew/brew/commits/bd6fb88ddc72a0de4c27f65e60a82790a5d87f3f/etc/bash_completion.d/brew?browsing_rename_history=true&new_path=completions/bash/brew&original_branch=master), twice to where [it lives now upstream](https://github.com/Homebrew/brew/commits/master/completions/bash/brew?after=171fba9c847c4855a0f2b493787871dcd1601e78+244).
If I understood correctly, the file upstream is autogenerated based on the Ruby API, based on the [release notes for 3.0.0](https://brew.sh/2021/02/05/homebrew-3.0.0/#post)
"[Bash](https://github.com/Homebrew/brew/pull/10229), [fish](https://github.com/Homebrew/brew/pull/10431) and [zsh](https://github.com/Homebrew/brew/pull/10403) completions are generated automatically from the CLI::Parser DSL. This will ensure they are kept up-to-date."

Just sharing in case there's useful changes which can be pulled down if it would be of use (as I said, I don't use this functionality).

mistydemeo:
The autogeneration that exists in Homebrew won't really apply to Tigerbrew at this point because they depend on a lot of development work that hasn't been merged in here. Definitely worth looking at the generated completions for inspiration though!

jcgraybill:
It looks like a lot of upstream work done in 2015 and 2016 was already pulled down to Tigerbrew, happily. I found three fixes that look useful and relevant to me and could be applied to Tigerbrew:
* https://github.com/Homebrew/legacy-homebrew/pull/45175
* https://github.com/Homebrew/legacy-homebrew/pull/45086
* https://github.com/Homebrew/brew/commit/09c04aa7155ee567f58e2b2285868bb81ce973c9

jcgraybill:
I'm also happy to break 👆 into three separate commits. I realize it's not the tidiest thing to bundle them into one.

sevan:
It should be ok as-is, just the one review item, do you want to switch to `$BASH_VERSINFO` instead of processing `$BASH_VERSION`?

PR #1371, opened 2025-05-20
arm: update urls

jcgraybill:
The source files for `arm` have moved, and the homepage is no longer current. Update both of these to new URLs. Note the checksum for the source tarball matches with the new URL.

In addition, `arm` requires the brew version of `python` on Tiger, so adding that dependency. 

Tested on:
* Tiger PowerPC & Intel
* Leopard PowerPC & Intel
* Snow Leopard Intel

PR #1376, opened 2025-05-22
arm: remove formula

jcgraybill:
Here's a PR for the alternative approach mentioned in https://github.com/mistydemeo/tigerbrew/pull/1371: removing the `arm` formula altogether and instructing users to install `nyx` from PyPI. Feel free to merge whichever approach you like most.

Tested on Snow Leopard / Intel.


mistydemeo (approved):
Looks good to me. Thank you!

Library/Homebrew/blacklist.rb
```diff
@@ -81,5 +81,9 @@ def blacklisted?(name)
     A binary installer is available:
       https://www.haskell.org/platform/mac.html
     EOS
+  when "arm" then <<-EOS.undent
+    Arm has been renamed to Nyx (https://nyx.torproject.org/) and is available through PyPI.
```
mistydemeo, line 85: Nitpick: looks like `arm` prefers the undercover "a" in its name.

jcgraybill: Good catch - updated.

PR #1377, opened 2025-05-23
frotz 2.44

jcgraybill:
I'd like to submit a formula for [Frotz](https://davidgriffith.gitlab.io/frotz/), the interpreter for text adventure games.

Version 2.50 of Frotz and beyond need `_strndup`, which I think requires Mac OS X 10.7 Lion or later - but please correct me if you know a workaround for this. So version 2.44 looks to be the latest version that can be built on Mac OS X 10.4 through 10.6. 

Modern Frotz comes in three flavors - a curses-based variant, a 'dumb' variant, and an SDL/graphical variant. I'm not able to figure out the SDL variant right now, but I thought it'd be great to start with the other two while I figure out whether this version of SDL frotz is in fact buildable.

Tested on Tiger and Leopard PowerPC; and Tiger, Leopard, and Snow Leopard Intel.

### DIFF
--- a/Library/Formula/frotz.rb
+++ b/Library/Formula/frotz.rb
@@ -5,7 +5,6 @@ class Frotz < Formula
   sha256 "dbb5eb3bc95275dcb984c4bdbaea58bc1f1b085b20092ce6e86d9f0bf3ba858f"
 
   depends_on "pkg-config" => :build
-  depends_on "ncurses"
 
   resource "testdata" do
     url "https://gitlab.com/DavidGriffith/frotz/-/raw/2.53/src/test/etude/etude.z5"
--- a/Library/Formula/frotz.rb
+++ b/Library/Formula/frotz.rb
@@ -1,0 +1,29 @@
+class Frotz < Formula
+  desc "Interpreter for Infocom games and other Z-machine games"
+  homepage "https://davidgriffith.gitlab.io/frotz/"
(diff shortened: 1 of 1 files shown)


sevan:
Thanks for the pull request.
Managed to build it just fine with GCC 4.0.1 on Tiger/powerpc. However, you depend on `ncurses` yet `bin/frotz` appears to link to the OS supplied version.

Your call whether to drop the dependency or fix the build.

```
$ otool -L bin/frotz
bin/frotz:
	/usr/lib/libncurses.5.4.dylib (compatibility version 5.4.0, current version 5.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 88.1.12)
```

sevan:
Regarding missing functionality such as `strndup()` it doesn't apply in this case but sometimes software carries implementations for fixing the build on other operating systems so if you're lucky you can just use that.
Other option is to provide an implementation from a library. I have landed versions of older software like you are doing here until that mythical library is sorted (working on it, just don't have anything to offer at the moment).

jcgraybill:
Thanks for catching the `ncurses` issue. I just double checked that indeed the `frotz` linked against the system `ncurses` passes all the tests in `etude.z5`, so there's no need for an additional `ncurses` dependency. I've removed it.

I bumped into the `strndup` wall with later versions of `erlang` too, and I see a couple other formulas that reference it. The good news is that `frotz` 2.44 is compatible with just as many games as the latest version of `frotz`, and there aren't scary security issues in it or anything.

PR #1379, opened 2025-05-26
frotz: install man pages

jcgraybill:
The Makefile for Frotz needs to have a variable called MAN_PREFIX defined in order to install man pages in the expected location for Tigerbrew. `frotz` now passes `brew audit --strict --online`.

Would a change like this warrant adding a `revision` number to the formula?

### DIFF
--- a/Library/Formula/frotz.rb
+++ b/Library/Formula/frotz.rb
@@ -13,6 +13,7 @@ class Frotz < Formula
 
   def install
     inreplace "Makefile", "PREFIX = /usr/local", "PREFIX = #{prefix}"
+    inreplace "Makefile", "MAN_PREFIX = $(PREFIX)", "MAN_PREFIX = #{man}/.."
     system "make", "all"
     system "make", "install"
     system "make", "install_dumb"


sevan:
> Would a change like this warrant adding a `revision` number to the formula?

Yes, because it's a fix.
I tend to omit the revbump if it doesn't apply/benefit powerpc users on larger programs to save a lengthy recompile for no benefit. e.g build fix for Snow Leopard.

jcgraybill:
You bet. I could use a pointer. I see from https://github.com/mistydemeo/tigerbrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#patches that the string HOMEBREW_PREFIX will get expanded out to the homebrew prefix before the patch is applied, but I'm not clear how to put the correct values for #{prefix}, #{etc}, and #{share} into a patch. If I understand right, these are supposed to be Cellar paths - e.g. `/usr/local/Cellar/frotz/2.44`. Is there an example of a formula that does this that you'd suggest I look at?

mistydemeo:
The other thing I'll mention is that make variables can typically be defined on the commandline when calling `make`. Rather than patching the makefile, it may be enough to do:

```
system "make", "PREFIX=#{prefix}" # and so on
```

sevan:
> The other thing I'll mention is that make variables can typically be defined on the commandline when calling `make`. Rather than patching the makefile, it may be enough to do:
> 
> ```
> system "make", "PREFIX=#{prefix}" # and so on
> ```

This is true, but you'd have to do it for each invocation of `make`.

jcgraybill:
> > This is true, but you'd have to do it for each invocation of `make`.
> 
> Unless you only invoke `make` once 🤔🤯
> 
> jcgraybill, you don't have to make this change, I'm just sharing for fun: `system "make", "all", "install", "install_dumb", "PREFIX = #{prefix}", "CONFIG_DIR = #{etc}", "MAN_PREFIX = #{share}"` works.

Hard core!

jcgraybill:
Thanks!

PR #1415, opened 2025-07-25
spidermonkey: require mac os x leopard

jcgraybill:
While working on `couchdb` builds for https://github.com/mistydemeo/tigerbrew/pull/1414 I confirmed that no version of `spidermonkey` will build on Mac OS X Tiger. I'd propose labeling the `spidermonkey` formula as requiring `:leopard` or greater to clarify this for users.


mistydemeo (approved):
Makes sense, thank you!

Issue #918, opened 2023-08-11
Cannot install Tigerbrew on x86 Leopard due to TLS problem


Amatarou-SUN:
I'm trying to install Tigerbrew on an x86 Leopard, but I get a TLS error and can't install.

[Error message]

==> Downloading and Installing Homebrew...
curl: (35) error:1409442E:SSL routines:SSL3_READ_BYTES:tlsv1 alert protocol version
Failed during: /usr/bin/curl -Lk -1 https://archive.org/download/tigerbrew/portable-curl-7.58.0.tiger_i386.bottle.tar.gz -o /tmp/curl.tar.gz

mistydemeo:
I think I've fixed this! You may not be able to `brew update` safely because of the bug, but if you edit `/usr/local/Library/Homebrew/cmd/vendor-install.sh` to replace the `https` URLs with `http`, it should work. Can you try that?

Amatarou-SUN:
Wow!!! Working so fine!!! thank you for your fix!!!

mistydemeo:
You're welcome! Please let me know if you have any other issues.

gothickitty93:
Sorry to necro this bug, but I think it's still present. I used this fix on a PPC G4 Xserve running Mac OS X 10.5.9 Sorbet Leopard, and it worked. Perhaps this edit should be the default?

sevan:
@gothickitty93 Are you working from an up to date Tigerbrew tree? [Misty commited the fix at the time](https://github.com/mistydemeo/tigerbrew/commit/3eb8081bf9f713feca8ee898145d4ee903186dfb) so it should be the default.

gothickitty93:
[snippet.log](https://github.com/user-attachments/files/21207675/snippet.log)

> [@gothickitty93](https://github.com/gothickitty93) Are you working from an up to date Tigerbrew tree? [Misty commited the fix at the time](https://github.com/mistydemeo/tigerbrew/commit/3eb8081bf9f713feca8ee898145d4ee903186dfb) so it should be the default.

I hope so. I followed the install instructions on the main page to download the linked 'install' file and run it with ruby. Are these instructions no longer accurate?

It's strange, though. The fix is clearly committed (thanks for the link btw). I've attached a snippet from my console that shows both before and after the edit.

mistydemeo:
[...]

PR #1417, opened 2025-08-11
bootstrap curl: use HTTP-only mirror

mistydemeo:
This mirror allows HTTP-only connections and should be more stable for installations than archive.org. Once we have this, we can then use it to fetch other packages from HTTPS-requiring sources.

This should fix remaining issues like the ones discussed in #918.

This mirror was graciously provided by @XerTheSquirrel.

cc @sevan

### DIFF
--- a/install
+++ b/install
@@ -5,9 +5,9 @@
 HOMEBREW_PREFIX = '/usr/local'
 
 CURL_VERSION    = "7.58.0"
-CURL_PPC_URL    = "http://archive.org/download/tigerbrew/portable-curl-#{CURL_VERSION}.tiger_g3.bottle.tar.gz"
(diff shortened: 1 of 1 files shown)


sevan:
Sorry about the delay, was offline for a bit.
Checked the dual stacked host is serving a file listing, and and the `portable-curl` archives are fetchable.
LGTM. Thanks for the mirror @XerTheSquirrel.

sevan:
@mistydemeo I think we need to try again (my git foo is too weak to correct), this went into the go branch by mistake.

mistydemeo:
That's correct actually! The `go` branch contains the installer; I'll update the `vendor-install` script separately.

sevan:
Ah, ok.
Sorry about the noise :)

PR #1435, opened 2025-10-02
cmake: fix building against curl 8.16.0

### LINKED ISSUE #1434: cmake failed to build on 10.5
Diagnostic info per the troubleshooting page linked below:

[gist](https://gist.github.com/betonmischer86/d5fcfbfb3fc839c4da9443f09a0a760e)

sevan:
Resolves #1434


Library/Formula/cmake.rb
```diff
@@ -3,14 +3,18 @@ class Cmake < Formula
   homepage "http://www.cmake.org/"
   url "https://cmake.org/files/v3.9/cmake-3.9.6.tar.gz"
   sha256 "7410851a783a41b521214ad987bb534a7e4a65e059651a2514e6ebfc8f46b218"
-  revision 1
+  revision 2
```
mistydemeo, line 6: Do we need to revbump it? This is a build error, and I'm not clear if this breaks using curl at runtime.

PR #1440, opened 2025-10-10
ranger 1.9.4

betonmischer86:
Bump version to 1.9.4 and rewrite the formula. Make Python 3 a dependency. Although Python 2.6 should be sufficient, I found it breaking plugins such as `ranger-devicons`, and some Linux distros require Python 3 for 1.9.4 anyway. One question remains.

A python3 shebang is currently patched in with this line:
`s.gsub! "#!/usr/bin/python", "#!#{Python3Requirement.new.which_python}"`

We end up with:
`#!/usr/local/opt/python3/bin/python3.10`

Instead of pointing to a specific version of Python 3, Is there a reason not to use the following? And what's the best way to arrive at that path programmatically in a formula (in case a non-standard Brew prefix is used)?
`#!/usr/local/opt/python3/bin/python3`

Tested on PowerPC G5 with Sorbet Leopard. More extensive testing is welcome, as I'm not a power user of Ranger or lf.

### DIFF
--- a/Library/Formula/ranger.rb
+++ b/Library/Formula/ranger.rb
@@ -1,34 +1,90 @@
 class Ranger < Formula
   desc "File browser"
-  homepage "http://ranger.nongnu.org/"
-  url "http://ranger.nongnu.org/ranger-1.7.1.tar.gz"
-  sha256 "f8b06135165142508ae7ec22ab2c95f6e51b4018c645d11226086d4c45b7df86"
+  homepage "https://ranger.github.io"
+  url "https://ranger.github.io/ranger-1.9.4.tar.gz"
+  sha256 "7ad75e0d1b29087335fbb1691b05a800f777f4ec9cba84faa19355075d7f0f89"
+  license "GPL-3.0-or-later"
+  head "https://github.com/ranger/ranger.git", branch: "master"
 
-  head "git://git.savannah.nongnu.org/ranger.git"
+  depends_on :python3
 
-  bottle do
-    cellar :any_skip_relocation
-    sha256 "b71c72d76937f8ae47761ba3c642eaef9e121a1b9fc5173d56a42b321f80e9ec" => :el_capitan
-    sha256 "5f0901ba96568d3f8ba713bbb39939538cbcdd1dffa06beee98c5cce13405d20" => :yosemite
-    sha256 "97bf46731d0b5cc20adbd3bd6052ca36508cb3b6deccb80c42cd5cc1ab5d83bc" => :mavericks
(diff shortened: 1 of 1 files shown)


sevan:
> We end up with: `#!/usr/local/opt/python3/bin/python3.10`
> 
> Instead of pointing to a specific version of Python 3, Is there a reason not to use the following? And what's the best way to arrive at that path programmatically in a formula (in case a non-standard Brew prefix is used)? `#!/usr/local/opt/python3/bin/python3`

It shouldn't be a problem, the substitution should work regardless of the non-standard prefix.

> Tested on PowerPC G5 with Sorbet Leopard. More extensive testing is welcome, as I'm not a power user of Ranger or lf.

Will take it for a spin on more releases and report back over the next few days.

mistydemeo:
> And what's the best way to arrive at that path programmatically in a formula (in case a non-standard Brew prefix is used)?

For reference - you can ask for the path to any formula by instantiating a formula object and calling methods on it. For example, `Formula["python3"].opt_prefix` will get you `/usr/local/opt/python3`, with the right Homebrew prefix included. There are also `opt_*` methods for specific subdirectories, like `opt_bin` - so you can do `"#{Formula["python3"].opt_bin}/python3"` to get a string with `/usr/local/opt/python3/bin/python3`.

sevan:
I tried it briefly on Tiger, didn't realise it was such a quick install of copying in a script.
It looks like at least on Tiger that it wants to be run from the desktop, where as I was running it via SSH initially. (not a problem)
I'm not sure if it's possible to just flip to using Python 3 without taking care of the options. I only looked at the first one which was libcaca :) and that's Python 2 based, tried upgrading it to a new release but I couldn't get it to pick up our Python 3 (tbh I didn't try too hard). Have you used many of the options which are now available in this formula, @betonmischer86 ?

betonmischer86:
@sevan, thank you for giving it a try. In the optional dependencies, I've included all of the formulae Ranger can integrate with, despite some of them being broken (e.g., #1444). Since none of those are mandatory, my reasoning was that any issues with installing them will have to be dealt with individually instead of removing problematic ones from Ranger's formula. IMO, it's even beneficial to keep outright broken options (like `w3m`) as a reminder to eventually fix them. Otherwise someone will have to keep track of what's being updated and make ongoing changing to the Ranger formula.

I can confirm at least those are working as expected with Ranger 1.9.4 on Sorbet: `atool`, `jq`, `exif-tool`, `highlight`. Also, tweaks to the Ranger config post-install might be required for some of the preview tools, but that I'd leave to the user.

PR #1469, opened 2025-12-05
gnu-sed: Add patch to fix #pragma diagnostics error when building with GCC 4.2

matthewyang204:
This adds a patch to remove unneeded GCC diagnostics pragma, which has been causing build errors in the likes of `#pragma GCC diagnostics in functions not support`.

### DIFF
--- a/Library/Formula/gnu-sed.rb
+++ b/Library/Formula/gnu-sed.rb
@@ -15,6 +15,11 @@ class GnuSed < Formula
 
   option "with-default-names", "Do not prepend 'g' to the binary"
 
+  patch :p1 do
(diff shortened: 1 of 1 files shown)


matthewyang204:
Update: Build from src succeeded on my iMac G5 running Leopard with the new modified formula.

sevan:
Thanks for the patch.
As usual, GCC 4.0 cares less about things and compiles just fine out of the box on Tiger with GCC 4.0.1.
But I can reproduce your issue on Tiger by building with GCC 4.2

```
sed/compile.c: In function 'bad_command':
sed/compile.c:206: error: #pragma GCC diagnostic not allowed inside functions
sed/compile.c:207: error: #pragma GCC diagnostic not allowed inside functions
sed/compile.c:209: error: #pragma GCC diagnostic not allowed inside functions
```

Issue #1459, opened 2025-11-17
Uninstall Script


smnbldwn:
Hello,
I just posted this in the wrong place. Not really an issue, just a question.

Is there an uninstall script for Tigerbrew or can I use the standard Homebrew one?

Thanks.

sevan:
The [FAQ pointed to the Homebrew uninstall script](https://github.com/mistydemeo/tigerbrew/blob/master/share/doc/homebrew/FAQ.md#how-do-i-uninstall-homebrew) however the link is dead.
Did you use the install script originally to get setup or just clone the repo somewhere and build away?

mistydemeo:
Sorry about that! I think it'd be a good idea to import the uninstall script into this repo, but in the meantime I've updated the FAQ to point to an older commit in that repo that was current at the time that FAQ was written. It should work.

mistydemeo:
I've updated the FAQ to point at the right Homebrew one. That should work.

smnbldwn:
Thank you.
When I run the command, I see the following error messages:

`
curl: (35) error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol`

sevan:
That's probably running the matured version of curl bundled with your OS. You need to us the version of curl from Tigerbrew.

Does this work? (assuming you have tigerbrew installed in `/usr/local`

`ruby -e "$(/usr/local/opt/curl/bin/curl -fsSL https://raw.githubusercontent.com/Homebrew/install/d8aeae7efb14ed9e4f7934b4782f04d5305562cc/uninstall)"`

smnbldwn:
Thanks Sevan.
When I use this command. I get a > symbol and a flashing cursor as if I have opened Ruby, but no more. 
Wait, my brain is switching on. I didn't have Curl installed... I am installing it now.

sevan:
> When I use this command. I get a > symbol and a flashing cursor as if I have opened Ruby, but no more.

bleh, appologies. Missed the close bracket.

smnbldwn:
After installing fresh Homebrew Curl:

```
[...]

PR #1471, opened 2025-12-06
docs: use vendored ruby & modern curl

sevan:
curl included with the OS most definitely will be unable to fetch the script due to ancient OpenSSL, and the uninstall script needs a newer version of ruby than is provided at the very least in Leopard and Tiger.

uninstall script uses String#start_with?
String#start_with? showed up in Ruby 1.8.7.
Ruby shipped in Leopard is 1.8.6

Discovered in issue #1459


sevan:
Another option is to carry a modified version of the uninstall script which works with Ruby 1.8.6.

PR #1482, opened 2025-12-21
vim 9.1.2006

sevan:
I think TASK_DEFAULT_APPLICATION showed up in Lion but can't test at the moment, so just drop the code for now. vim wil behave as before this snippet of code was introduced into the vim codebase.

Tested on Tiger PowerPC (G3) with GCC 4.0.1
Built on 600Mhz iBook G3 running Tiger using GCC 4.0.1 in 21 minutes.

### DIFF
--- a/Library/Formula/vim.rb
+++ b/Library/Formula/vim.rb
@@ -2,15 +2,16 @@ class Vim < Formula
   desc "Vi \"workalike\" with many additional features"
   homepage "http://www.vim.org/"
   # *** Vim should be updated no more than once every 7 days ***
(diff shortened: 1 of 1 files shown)


mistydemeo:
> I think TASK_DEFAULT_APPLICATION showed up in Lion but can't test at the moment

Yep, I think you're right: https://developer.apple.com/documentation/kernel/task_role_t/task_default_application?changes=_9

sevan:
Thanks for checking, I have a vague memory that it shows up in 10.6.8 as well but not sure if it's just to fill the space for backwards compatibility or they actually added the functionality. Will test & update things.

PR #1485, opened 2025-12-22
add G3 bottles for several packages

mistydemeo:
These bottles were built by Sevan; I've just handled the uploads.


sevan:
Thanks, Misty :)

matthewyang204:
@mistydemeo I also have a few Leopard G5 bottles if you want them

Issue #775, opened 2021-11-08
This won't work for 10.6 PowerPC, right?


barracuda156:
I am looking for alternative of MacPorts, which does build on 10.6 PPC, but some ports fail to build.
I am specifically talking about 10.6 PowerPC, not Intel. I have 10A190 build installed and running.

mistydemeo:
10.6 PowerPC? I had no idea it existed! It hasn't been tested but I'd absolutely welcome any reports, and try to fix anything that's broken.

LukeUsher:
> 10A190

There's no retail/final 10.6, 10A190 is a beta from before they removed PowerPC support, that was handed out at some Apple Developer Conference events.

Maltettt:
I‘ve tried using Tigerbrew on the current 10.6 PPC build (Alpha Testing: Version 5), but brew tries to install package versions for the wrong CPU type (Bad CPU type in executable) for software that works fine on 10.5 PPC.

sevan:
What command did you run & which package was it installing?

sevan:
I wonder if commenting out line [28 & 29](https://github.com/mistydemeo/tigerbrew/blob/9216073492b151fd87d111adb410bf2dbaf47d72/Library/Homebrew/bottles.rb#L28C1-L29C64) in `Library/Homebrew/bottles.rb` will prevent it from making the wrong selection.

Maltettt:
I’ve tried of running brew doctor. Just now I‘ve realised, that the bad executable is actually clang. I‘ve tried running 10.6 PPC with XCode 3.2.6 and 3.1.4 and had no luck correcting that issue for now.

barracuda156:
> I’ve tried of running brew doctor. Just now I‘ve realised, that the bad executable is actually clang. I‘ve tried running 10.6 PPC with XCode 3.2.6 and 3.1.4 and had no luck correcting that issue for now.

1. Please do not use Leopard Xcode on 10.6.x, do yourself a favor.
2. Xcode 3.2.6 needs a number of executables rebuilt/replaced/substituted, because not everything has ppc slices. The easiest way which is acceptable is to pull over stuff from 10a190 installed on a drive or partition. Not advised, you get poor man’s Xcode. The proper alternatives are either to build from Apple sources (manually or via darwinbuild) or install something newer which works for the purpose (you only need a handful of CLI tools).
[...]

PR #1495, opened 2026-01-07
bottles: support Snow Leopard PPC

mistydemeo:
Little bit messy, but it'll work.

For historical reasons, :snow_leopard specifically means 64-bit Intel and :snow_leopard_32 means 32-bit Intel. This keeps support for those two with their existing semantics while also adding support for Snow Leopard PPC.

refs #775.


sevan:
@mistydemeo should be do this in reverse in that if there's no regression on Intel build of Snow Leopard, we merge it? I can test the Intel build, I don't have the PowerPC build.

PR #1501, opened 2026-02-13
x11: fix install message for Tiger/Leopard

### LINKED ISSUE #1499: Cannot install XQuartz on Mac OS 10.4
I was trying to install curl like this: "brew install curl". I then saw an error message stating "tcl-tk: XQuartz is required to install this formula."

The suggestion given is to run this command: "brew install Caskroom/cask/xquartz". When I run this command I see this error: "fatal: unable to access 'https://github.com/Caskroom/homebrew-cask/': Unkown SSL protocol error in connection to github.com: -9836".

My info:
Mac OS 10.4.11
Tigerbrew 0.9.5
PowerPC

mistydemeo:
The previous message doesn't work for Tigerbrew; we should just tell users to get it from the XQuartz site or their Mac OS X DVD, as appropriate. I've also added messages for specific OS versions that can't use the latest release.

cc @sevan - how do you feel about this messaging? Any other versions we want in?

Fixes #1499.

### DIFF
--- a/Library/Homebrew/requirements/x11_requirement.rb
+++ b/Library/Homebrew/requirements/x11_requirement.rb
@@ -5,7 +5,6 @@ class X11Requirement < Requirement
   attr_reader :min_version
 
   fatal true
-  cask "xquartz"
   download "https://xquartz.macosforge.org"
 
   env { ENV.x11 }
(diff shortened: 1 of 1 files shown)


sevan:
All good so far, just one addition. The case for Snow Leopard, should be extended to include & up to Mountain Lion.
XQuartz requires Mavericks as a minimum version for versions beyond 2.7.11
"[XQuartz 2.7.11](https://www.xquartz.org/releases/XQuartz-2.7.11.html) - 2016-10-29 - Last release to support Snow Leopard, Lion, and Mountain Lion"

mistydemeo:
Good call, done.

sevan:
LGTM

PR #1502, opened 2026-02-13
requirement: remove cask support

mistydemeo:
Cask doesn't support the Mac versions we do, and even if it did it wouldn't have casks that work on these OS releases. We should stop recommending casks as ways to install the software from these requirements.

X11 was previously removed in #1501, but these remaining ones would still recommend the user install a cask. All of them can be satisfied by Tigerbrew except Java, which we can eventually find a better recommendation for.

cc @sevan


sevan:
Makes sense to clean it up.

PR #1504, opened 2026-02-18
requirement: remove lingering cask reference

sevan:
Now that cask support has been removed.

Apologies, missed it as part of the #1502 review.


mistydemeo:
Oops, thanks! I missed that.

PR #1506, opened 2026-02-22
uninstaller: import from upstream, fix when ~/Applications missing

### LINKED ISSUE #1505: Uninstall script fails if ~/Applications doesn't exist
I tried this:
```
/usr/local/Library/Homebrew/vendor/portable-ruby/2.3.3/bin/ruby -e "$(/usr/local/opt/curl/bin/curl -fsSL https://raw.githubusercontent.com/Homebrew/install/d8aeae7efb14ed9e4f7934b4782f04d5305562cc/uninstall)"
```

For some reason, my Leopard user does not have the ~/Applications folder so the uninstall script fails. I manually added the ~/Applications folder so the script would succeed.

The FAQ says the uninstall command is this:
```
/usr/local/Library/Homebrew/vendor/portable-ruby/current/bin/ruby -e "$(/usr/local/Library/Homebrew/vendor/portable-curl/current/bin/curl -fsSL https://raw.githubusercontent.com/Homebrew/install/d8aeae7efb14ed9e4f7934b4782f04d5305562cc/uninstall)"
```

but I don't think it matters since they both use the same script.

mistydemeo:
We'd still been a copy of the upstream uninstall script as it was at https://github.com/Homebrew/install/commit/d8aeae7efb14ed9e4f7934b4782f04d5305562cc. This swaps over to a local copy of the script and modifies it to handle a missing `~/Applications`, as had been reported to us.

Fixes #1505.

### DIFF
--- a/uninstall
+++ b/uninstall
@@ -127,7 +127,10 @@ if /darwin/i === RUBY_PLATFORM
     /Applications
     #{ENV["HOME"]}/Applications
   ].map do |p|
-    Pathname.new(p).children.select do |app|
+    path = Pathname.new(p)
+    next unless path.exist?
+
+    path.children.select do |app|
       app.resolved_path.to_s.start_with? HOMEBREW_CELLAR.to_s
     end
   end.flatten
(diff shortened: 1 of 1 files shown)


mistydemeo:
cc @sevan - mind if I merge this and #1507?

sevan:
@mistydemeo sorry, I missed it. Go for it.

PR #1520, opened 2026-05-05
LICENSE.txt: Sync with upstream

sevan:
Mainly to obtain the ammendment to handling date.
This new revision referes to THE COPYRIGHT HOLDERS AND CONTRIBUTORS, instead of solely THE AUTHOR.

Obtained from https://github.com/Homebrew/homebrew-core/commit/40dd66b71ac6efa62e647f695c515dd9b79dc1ab

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,22 +1,25 @@
-Copyright 2009-2015 Homebrew contributors.
+BSD 2-Clause License
+
+Copyright (c) 2009-present, Homebrew contributors
+All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions
-are met:
+modification, are permitted provided that the following conditions are met:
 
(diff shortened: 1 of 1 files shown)


sevan:
@mistydemeo ok to merge?

LICENSE.txt
```diff
@@ -1,22 +1,25 @@
-Copyright 2009-2015 Homebrew contributors.
+BSD 2-Clause License
+
+Copyright (c) 2009-present, Homebrew contributors
```
mistydemeo: I wonder if we want to make this "Homebrew and Tigerbrew contributors" for us, since it covers both sets of developers.

mistydemeo (approved):
Sorry for the delay; I was out of town over the weekend and only just got back. Looking good!

LICENSE.txt
sevan: Good idea

PR #1529, opened 2026-06-01
giflib: update source url

dotsam:
The project seems to have reorganized their files, the download lives in a `giflib-5.x` subdirectory now.


sevan:
Thanks for the patch.

PR #1546, opened 2026-06-28
Handle no_optimization in superenv

sevan:
`ENV.no_optimization` has handled things in the `stdenv` where setting the default optimisation flags are skipped in `C/CXXFLAGS`. However, indirectly in the background, the `cc` wrapper script still sets `-Os` via the `$HOMEBREW_OPTIMIZATION_LEVEL` in the `superenv`. This causes issues on GCC builds from version 7 onwards where `g++` is always being invoked with `-Os` via `superenv` which generates a different compiler from the one which the build process does (`xgcc`) and when it comes to do the stage comparison, things fail.

With this change if `ENV.no_optimization` is set, `$HOMEBREW_OPTIMIZATION_LEVEL` is deleted from the build env, and the `cc` wrapper skips setting an optimisation level.

Fixes issue #554

### DIFF
--- a/Library/ENV/4.3/cc
+++ b/Library/ENV/4.3/cc
@@ -202,7 +202,7 @@ class Cmd
 
     args << "-pipe"
     args << "-w" unless configure?
-    args << "-#{ENV["HOMEBREW_OPTIMIZATION_LEVEL"]}"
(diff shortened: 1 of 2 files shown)


mistydemeo (approved):
Looks good! Definitely makes sense to add.

PR #1554, opened 2026-08-12
make: Avoid apple-gcc42 build 5553

### LINKED ISSUE #1513: Git 2.53 formula fails to compile on G4 10.4.11.  Gmake fails with memory errors
I've been trying to get through `brew update` but I've run into an issue with git I cannot seem to resolve.  Each time I try to install it `gmake install` fails withe the following messages (some of the numbers have changed around since the last time I checked):

> gmake(1303) malloc: *** error for object 0x1bf6c: pointer being reallocated was not allocated
> gmake(1303) malloc: *** set a breakpoint in szone_error to debug
> gmake: *** virtual memory exhausted

This would suggest to me that I don't have enough memory or I'm hampered by 32-bit addressing limitations here.  However, I don't really see a spike in memory usage beyond what I would expect and I've already successfully managed to build GCC 8.5 despite that having a significantly larger repo.

sevan:
Building with optimisation results in a make utiliy which fails with malloc errors when building Git.
It works if no_optimization is used but the problem is specific to apple-gcc42 build 5553, so I opted to block it off.

Current bottles were built with GCC 4.0.1 on tiger, so leaving them in place.

Fixes #1513


mistydemeo:
Very good catch, thanks!

PR #1556, opened 2026-08-22
Codeberg migration

mistydemeo:
This prepares us for a Codeberg migration. This is *not tested yet* (I haven't set up a copy of the repository yet!), but this replaces the key references in the code and documentation while also making code changes to the updater. The updater will now check if the remote is on GitHub and, if it is, replace the git origin with the new one. This should allow users to transparently move from GitHub to Codeberg just by running `brew update`. (Fun fact: this is leveraging the code that already existed to migrate from the old mxcl/homebrew name to Homebrew/homebrew, just with minor tweaks to get it working with a non-GitHub host.) I've also updated this migrate us from a `master` branch on GitHub to a `main` branch on Codeberg.

One thing I haven't yet done is rewrite the GitHub API client code that handled things like searching for open issues. That will need some attention.


sevan:
Looks good so far. `--bugurl` in `gcc.rb`, `gcc44.rb`, `gcc45.rb`, `gcc46.rb` need switching too.
Patch fixes that:
[gcc-bugurl-update.txt](https://github.com/user-attachments/files/31350065/gcc-bugurl-update.txt)

sevan:
There's 64 references to github gists for patches in formulas. Leave those for a separate thing to migrate?

sevan (approved):

sevan:
> There's 64 references to github gists for patches in formulas. Leave those for a separate thing to migrate?

Happy to handle that after the migration. I'll merge them into the formula files so everything is self contained.

mistydemeo:
Sounds great, thank you!

PR #1557, opened 2026-08-30
os/mac: don't crash when /usr/sbin/pkgutil is missing (Tiger)

watermark-hd:
On Mac OS X 10.4 there is no /usr/sbin/pkgutil, so MacOS.pkgutil_info raises
Errno::ENOENT from Utils.popen_read. It's reached during the post-install
receipt phase, so every `brew install` that gets as far as `make install`
dies with:

    Error: No such file or directory - /usr/sbin/pkgutil

Guarding the call with File.exist? and returning "" is enough: the callers
(Xcode CLT detection, XQuartz.version) already handle an empty result.

Tested on 10.4.11 / PowerPC: with this change, `brew install gcc`, openssl3,
libffi, sqlite, readline, gdbm and a from-source CPython 3.12 all complete
`make install`. Without it they all fail at the same point.


sevan:
Thanks for the patch. What version of Xcode are you running on OS X 10.4.11?
It shouldn't be an issue on version 2.5 which was the last release to support Tiger.

watermark-hd:
Just checked on the actual hardware: gcc -v reports gcc version 4.0.1 (Apple Computer, Inc. build 5370). 
That build number matches Xcode 2.5 (LLVM's own docs reference this exact "Xcode 2.5 / gcc 4.0.1 Apple Build 5370" pairing), so yes — 2.5, the last one that supports Tiger, same as you said.

mistydemeo:
Can you show me the full log where you encountered it? We have guards around the places calling the helper that intended to ensure we weren't calling it on OSs where it's not available, but we clearly missed a spot. I think we want to ensure we handle that properly so that we end up relying on null data coming out of this.

mistydemeo:
It looks like we call it in two places:

* In `Xcode::CLT.detect_version`, where we skip it if Xcode is installed and is older than Xcode 3: https://github.com/mistydemeo/tigerbrew/blob/f16f677a466434fc746138a40b4311beb95a85c2/Library/Homebrew/os/mac/xcode.rb#L206-L208
* In `MacOS::XQuartz.version_from_pkgutil`: https://github.com/mistydemeo/tigerbrew/blob/f16f677a466434fc746138a40b4311beb95a85c2/Library/Homebrew/os/mac/xquartz.rb#L82C11-L85 Which we call as a fallback if we didn't find a version from `mdls` and didn't guess a version from the path: https://github.com/mistydemeo/tigerbrew/blob/f16f677a466434fc746138a40b4311beb95a85c2/Library/Homebrew/os/mac/xquartz.rb#L36-L42

Do you have XQuartz/X11.app installed? I wonder if this issue only happens if you don't, and we didn't properly skip checking this version in this case.

watermark-hd:
No, XQuartz / X11.app is not installed here — plain Tiger 10.4.11 / PowerPC, `X11: N/A` in `brew config`. So the `XQuartz.version_from_pkgutil` path is not involved.

It's the **CLT path**, and it only triggers from the build-failure dump:
`BuildError#dump` -> `Homebrew.dump_verbose_config` -> `MacOS::CLT.installed?` -> `CLT.detect_version`.

Why the existing guard misses it: `detect_version` early-returns only when `MacOS::Xcode.installed? && MacOS::Xcode.version < "3.0"`. On Tiger there is no clang, so `MacOS::Xcode.version` returns the `"dunno"` sentinel (the `when 0 then "dunno"` branch in `uncached_version`), and `"dunno" < "3.0"` is `false` as a string comparison. So it falls through to `MacOS.pkgutil_info` -> `Utils.popen_read("/usr/sbin/pkgutil", ...)` -> `Errno::ENOENT`. (Real Xcode here is 2.5 at `/Developer`, gcc 4.0.1 build 5370.)

The side effect is that it hides the real build error. In the log below the actual failure is the bottle relocation dying on Tiger's old `install_name_tool`, then `configure` reporting "C compiler cannot create executables" — and the `pkgutil` error is printed on top of that:

~~~
==> Pouring gmp-6.3.0_1.tiger_altivec.bottle.tar.gz
/usr/bin/install_name_tool: object: /usr/local/Cellar/gmp/6.3.0_1/lib/libgmpxx.4.dylib malformed object (unknown load command 11)
Error: Failure while executing: /usr/bin/install_name_tool -id /usr/local/opt/gmp/lib/libgmpxx.4.dylib /usr/local/Cellar/gmp/6.3.0_1/lib/libgmpxx.4.dylib
Warning: Bottle installation failed: building from source.
==> Installing dependencies for gmp: xz
==> Installing gmp dependency: xz
==> Pouring xz-5.8.1.tiger_altivec.bottle.tar.gz
/usr/bin/install_name_tool: object: /usr/local/Cellar/xz/5.8.1/bin/xz malformed object (unknown load command 11)
Error: Failure while executing: /usr/bin/install_name_tool -change @@HOMEBREW_CELLAR@@/xz/5.8.1/lib/liblzma.5.dylib /usr/local/Cellar/xz/5.8.1/lib/liblzma.5.dylib /usr/local/Cellar/xz/5.8.1/bin/xz
Warning: Bottle installation failed: building from source.
==> Downloading https://downloads.sourceforge.net/project/lzmautils/xz-5.8.1.tar.gz
==> ./configure --disable-silent-rules --prefix=/usr/local/Cellar/xz/5.8.1
Last 15 lines from /Users/watermark/Library/Logs/Homebrew/xz/01.configure:
checking for gcc... /usr/bin/gcc-4.0
checking whether the C compiler works... no
configure: error: in '/private/tmp/xz20260829-19716-m7p1mp/xz-5.8.1':
configure: error: C compiler cannot create executables
See 'config.log' for more details
Error: No such file or directory - /usr/sbin/pkgutil
~~~

Possible fixes: the `File.exist?` guard in this PR, or gate `CLT.detect_version` on `MacOS.version < :leopard`, or make `Utils.popen` call `exit!` in the child when `exec` fails.

